PR #6, opened 2022-10-13
Add CI/CD

Sxnan:

Add CI that runs tests on every pull request and push in the future. Add CD that builds and push the nightly wheel of Feathub to PYPI daily.

## Brief change log

- Add CI that runs tests on every pull request and push in the future. 
- Add CD that builds and push the nightly wheel of Feathub to PYPI daily.
- Fix some tests that assume the time zone.

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/examples/streaming_average_flink_cli.py
+++ b/python/feathub/examples/streaming_average_flink_cli.py
@@ -84,7 +84,7 @@ def main() -> None:
     )
 
     sink = PrintSink()
-    client.materialize_features(feature_view, sink=sink, allow_overwrite=True).wait()
+    client.materialize_features(feature_view, sink=sink, allow_overwrite=True)
 
 
 if __name__ == "__main__":
--- a/tools/ci/run-examples.sh
+++ b/tools/ci/run-examples.sh
@@ -1,0 +1,61 @@
+#
+# Copyright 2022 The Feathub Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


python/feathub/examples/streaming_average_flink_cli.py
```diff
@@ -84,7 +84,7 @@ def main() -> None:
     )
 
     sink = PrintSink()
-    client.materialize_features(feature_view, sink=sink, allow_overwrite=True).wait()
+    client.materialize_features(feature_view, sink=sink, allow_overwrite=True)
```
lindong28, line 87: If we remove this `wait()`, can we still ensure that the `PrintSink()` will print all results before the program exits?

tools/ci/run-examples.sh
```diff
@@ -0,0 +1,61 @@
...
+echo "Running Flink processor cli mode quickstart."
+bash tools/cli-deps/build-cli-deps.sh
+./flink-"${FLINK_VERSION}"/bin/flink run \
+    --detach \
+    --python python/feathub/examples/streaming_average_flink_cli.py \
+    --pyFiles tools/cli-deps/deps.zip
+
+# Wait until there are at least 10 output from TM stdout.
+while true; do
+  if [ "$(cat ./flink-"${FLINK_VERSION}"/log/*taskexecutor*.out | wc -l)" -ge 10 ]; then
```
lindong28, line 53: Can we additionally check the content of the `*.out` file to ensure that `nyc_taxi_flink_session.py` has finished successfully? Maybe let `nyc_taxi_flink_session.py` print something like `Successfully finished...` in the output file?

lindong28 (commented):
Thanks for the PR! LGTM overall. Just left some minor comments.

tools/ci/run-examples.sh
Sxnan: Here we are checking the result of the streaming job streaming_average_flink_cli.py that is submitted by the `flink run` command. 

For nyc_taxi_flink_session.py, since it is a session job. If the python processor on the client side exit successfully, we know that the flink job finished without exception. I think it should be enough to ensure that our examples are not broken.

Sxnan:
@lindong28 Thank you for your comments. I have updated the PR and replied to your comments. Could you please take another look?

lindong28:
Thanks @Sxnan for the explanation! LGTM.

PR #8, opened 2022-10-18
Update Feathub expression

Sxnan:

This PR refactors the Feathub expression parser so that each processor can implement its own logic to evaluate the syntax tree. Introduce cast, logical operation, and keyword escape syntax to Feathub expression. 

## Brief change log

- Refactors the Feathub expression parser 
- Introduce cast, logical operation (and, or), and keyword escape syntax to Feathub expression
- Implement AST evaluator for Flink processor

  - The public API: yes

## Documentation


python/feathub/examples/nyc_taxi_flink_session.py
```diff
@@ -194,8 +194,8 @@ def build_features(client: FeathubClient) -> FeatureView:
     f_trip_time_duration = Feature(
         name="f_trip_time_duration",
         dtype=types.Int32,
-        transform="unix_timestamp(CAST(lpep_dropoff_datetime AS STRING)) - "
-        "unix_timestamp(CAST(lpep_pickup_datetime AS STRING))",
+        transform="UNIX_TIMESTAMP(CAST(lpep_dropoff_datetime AS STRING)) - "
```
lindong28, line 197: Could you double-check the difference between this file and `nyc_taxi.py` and make sure that they only diff in the client configuration?

And it is possible to put the common part as method(s) in `nyc_taxi.py` and let `nyc_taxi_flink_session.py` invoke the method(s) instead of duplicating the FeatHub feature definitions?

python/feathub/processors/local/ast_evaluator/ast_evaluator.py
```diff
@@ -0,0 +1,144 @@
...
+    CastOp,
+    GroupNode,
+)
+from feathub.processors.local.ast_evaluator.functions import get_predefined_function
+
+_TRUE_STRINGS = ("t", "true", "y", "yes", "1")
+_FALSE_STRINGS = ("f", "false", "n", "no", "0")
+
+
+class AstEvaluator(AbstractAstEvaluator):
```
lindong28, line 37: Would it be more readable to rename this class as `LocalAstEvaluator`? This would also be more consistent with the way we name `FlinkProcessor` and `LocalProcessor`.

If we make this change, please rename test class as appropriate.

python/feathub/processors/flink/ast_evaluator/ast_evaluator.py
```diff
@@ -0,0 +1,79 @@
...
+    FuncCallOp,
+    ValueNode,
+    CompareOp,
+    UminusOp,
+    BinaryOp,
+    GroupNode,
+)
+
+
+class AstEvaluator(AbstractAstEvaluator):
```
lindong28, line 31: Would it be more readable to rename this class as `FlinkAstEvaluator`? This would also be more consistent with the way we name `FlinkProcessor` and `LocalProcessor`.

If we make this change, please rename test class as appropriate.

lindong28 (commented):
Thanks for the PR. Looks pretty good! Left just some minor comments.

lindong28 (approved):
Thanks for the update. LGTM.

PR #15, opened 2022-10-25
Sliding window handles expired value and duplicate value

Sxnan:

**Motivation**

Currently, Flink sliding window only output result when a sliding window has at least one data point in the sliding window. When the sliding window becomes empty, the sliding window will not output anything to let the downstream know that the current sliding window becomes empty. Usually, it requires the external system to clean up the expired result periodically or to check the timestamp of the result to determine if it is still valid when using the result.
And the Flink sliding window outputs results at every step, even when the result doesn't change, which causes unnecessary usage of network IO.
In this PR, we want to improve the Feathub sliding window so that it can output a message when a sliding window becomes empty and skip the same sliding window output.

**Public API**

We introduce two global configurations to config the behavior of the sliding window.
- `sdk.sliding_feature_view.enable_empty_window_output`: Default to True. If it is True, when the sliding window becomes emtpy, it outputs zero value for aggregation function SUM and COUNT, and output None for other aggregation function. If it is False, the sliding window doesn't output anything when the sliding window becomes empty.
- `sdk.sliding_feature_view.skip_same_window_output`: Default to True. If it is True, the sliding feature view only outputs when the result of the sliding window changes. If it is False, the sliding feature view outputs at every step size even if the result of the sliding window doesn't change. 

By default, `enable_empty_window_output` is `True` and `skip_same_window_output` is `True` so that we get the best performance of the sliding window.
For users who prefer the Flink native sliding window behavior, where there is no output on an empty window and without skipping the same window output, they can set `enable_empty_window_output` to `False` and `skip_same_window_output` to `False`.
For users who want to get the sliding window result's last updated timestamp and the message when the result is expired, they can set `enable_empty_window_output` to `True` and `skip_same_window_output` to `False`.
Note that we prevent setting `skip_same_window_output` to `True` when `enable_empty_window_output` is `False` because, with this configuration, there is not enough information to determine if a result is expired or not. 

**Implementation**

In this PR, we introduce the `PostSlidingWindowKeyedProcessFunction` that Feathub will apply after the native Flink sliding window so that it can handle the expired row and skip the same sliding window result. If the sliding window has group-by-keys, the `PostSlidingWindowKeyedProcessFunction` requires storing one row for each key to the Flink state. If the sliding window doesn't have a group-by-key, the `PostSlidingWindowKeyedProcessFunction` only requires storing one row in the Flink state.

## Brief change log

- Introduce PostSlidingWindowKeyedProcessFunction that can handle expired values and skip the same window results
- Introduce SlidingWindowConfig with global config `sdk.sliding_feature_view.enable_empty_window_output` and `sdk.sliding_feature_view.skip_same_window_output`

  - The public API: yes

## Documentation

### DIFF
--- a/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PostSlidingWindowKeyedProcessFunction.java
+++ b/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PostSlidingWindowKeyedProcessFunction.java
@@ -1,0 +1,175 @@
+/*
+ * Copyright 2022 The Feathub Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.feathub.flink.udf;
+
+import org.apache.flink.api.common.state.MapState;
+import org.apache.flink.api.common.state.MapStateDescriptor;
+import org.apache.flink.api.common.typeinfo.TypeInformation;
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.common.typeutils.base.LongSerializer;
+import org.apache.flink.api.java.tuple.Tuple2;
+import org.apache.flink.api.java.typeutils.ResultTypeQueryable;
+import org.apache.flink.api.java.typeutils.runtime.TupleSerializer;
+import org.apache.flink.configuration.Configuration;
+import org.apache.flink.streaming.api.functions.KeyedProcessFunction;
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.catalog.Column;
+import org.apache.flink.table.catalog.ResolvedSchema;
+import org.apache.flink.table.runtime.typeutils.ExternalTypeInfo;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+
+import java.time.Instant;
+
+/**
+ * The KeyedProcessFunction that is used after the Flink SQL Sliding Window operation by Feathub.
+ *
+ * <p>The function output the rows only when the result of the sliding window changes if
+ * skipSameWindowOutput is true. And it handles the expired row with the given {@link
+ * PostSlidingWindowExpiredRowHandler}.
+ *
+ * <p>It assumes that the rows of each key are ordered by the row time, which should hold true after
+ * a sliding window.
+ */
+public class PostSlidingWindowKeyedProcessFunction extends KeyedProcessFunction<Row, Row, Row>
+        implements ResultTypeQueryable<Row> {
+
+    private final long windowStepSizeMs;
+    private final TypeSerializer<Row> keySerializer;
+    private final TypeSerializer<Row> rowTypeSerializer;
+    private final ExternalTypeInfo<Row> rowTypeInfo;
+
+    private final String rowTimeFieldName;
+    private final PostSlidingWindowExpiredRowHandler expiredRowHandler;
+    private final boolean skipSameWindowOutput;
+
+    private MapState<Row, Tuple2<Long, Row>> lastRowState;
+
+    public PostSlidingWindowKeyedProcessFunction(
+            ResolvedSchema schema,
+            long windowStepSizeMs,
+            String[] keyFieldNames,
+            String rowTimeFieldName,
+            PostSlidingWindowExpiredRowHandler expiredRowHandler,
+            boolean skipSameWindowOutput) {
+        this.windowStepSizeMs = windowStepSizeMs;
+        this.rowTimeFieldName = rowTimeFieldName;
+        this.expiredRowHandler = expiredRowHandler;
+        this.skipSameWindowOutput = skipSameWindowOutput;
+
+        DataTypes.Field[] keyFields = new DataTypes.Field[keyFieldNames.length];
+        for (int i = 0; i < keyFields.length; ++i) {
+            final String keyFieldName = keyFieldNames[i];
+            final Column keyCol =
+                    schema.getColumn(keyFieldName)
+                            .orElseThrow(
+                                    () ->
+                                            new RuntimeException(
+                                                    String.format(
+                                                            "The given key field %s doesn't exist.",
+                                                            keyFieldName)));
+            keyFields[i] = DataTypes.FIELD(keyCol.getName(), keyCol.getDataType());
+        }
+
+        this.keySerializer =
+                ExternalTypeInfo.<Row>of(DataTypes.ROW(keyFields)).createSerializer(null);
+        rowTypeInfo = ExternalTypeInfo.of(schema.toPhysicalRowDataType());
+        this.rowTypeSerializer = rowTypeInfo.createSerializer(null);
+    }
+
+    @Override
+    public void open(Configuration parameters) {
+        @SuppressWarnings({"unchecked", "rawtypes"})
+        final TupleSerializer<Tuple2<Long, Row>> tupleSerializer =
+                new TupleSerializer(
+                        Tuple2.class,
+                        new TypeSerializer[] {LongSerializer.INSTANCE, rowTypeSerializer});
+        lastRowState =
+                getRuntimeContext()
+                        .getMapState(
+                                new MapStateDescriptor<>(
+                                        "lastRowState", keySerializer, tupleSerializer));
+    }
+
+    @Override
+    public void processElement(
+            Row row, KeyedProcessFunction<Row, Row, Row>.Context ctx, Collector<Row> out)
+            throws Exception {
+        final Row currentKey = ctx.getCurrentKey();
+        final Tuple2<Long, Row> lastRowTuple = lastRowState.get(currentKey);
+        final Instant rowInstant = row.getFieldAs(rowTimeFieldName);
+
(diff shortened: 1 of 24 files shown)


Sxnan:

python/feathub/processors/flink/table_builder/sliding_window_utils.py
```diff
@@ -202,6 +212,15 @@ def evaluate_sliding_window_transform(
             ],
         )
 
+    if empty_window_output_enabled:
```
lindong28, line 215: In case `empty_window_output_enabled == false`, we still want to allow users to skip emitting the same window values, right?

Could you provide the pointer to the code that enables this choice?

python/feathub/processors/flink/table_builder/sliding_window_utils.py
```diff
@@ -108,16 +112,22 @@ def evaluate_sliding_window_transform(
...
     Evaluate the sliding window transforms on the given flink table and return the
     result table.
 
+
     :param t_env: The StreamTableEnvironment of the `flink_table`.
     :param flink_table: The input Flink table.
     :param window_descriptor: The descriptor of the sliding window.
     :param agg_descriptors: A list of descriptor that descriptor the aggregation to
                             perform.
+    :param deduplicate_output_enabled: Whether to deduplicate sliding window output.
```
lindong28, line 128: FeatHub users typically will not read this Python doc.

Should we document the semantics of these two configs in `sliding_window_transform.py`?

And should we add a TODO regarding retracting the value when the window becomes empty?

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/PostSlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,172 @@
...
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.catalog.Column;
+import org.apache.flink.table.catalog.ResolvedSchema;
+import org.apache.flink.table.runtime.typeutils.ExternalTypeInfo;
+import org.apache.flink.table.utils.DateTimeUtils;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+
+/**
+ * The KeyedProcessFunction that is used after the Flink SQL Sliding Window operation by Feathub.
```
lindong28, line 38: Can we consistently use `FeatHub` instead of `Feathub` in this PR? We can change the existing doc/code in a separate PR.

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/PostSlidingWindowExpiredRowHandler.java
```diff
@@ -0,0 +1,27 @@
...
+
+package com.alibaba.feathub.udf;
+
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+
+import java.io.Serializable;
+
+/** Interface for handling post sliding window expired row. */
+public interface PostSlidingWindowExpiredRowHandler extends Serializable {
```
lindong28, line 25: Can you explain another possible implementation of this interface? Maybe add a TODO here?

lindong28 (commented):

python/feathub/processors/flink/table_builder/sliding_window_utils.py
Sxnan: I think we prevent setting `enable_static_window_output` to `False`, which skip emitting the same window values when `enable_empty_window_output` is `False`. Because with this configuration, it is impossible to determine if a value is expired or not. It is checked when FeathubClient gets the sdk_config in config.py:55.

Sxnan:
@lindong28 Thanks for your review. I updated the PR. Could you please take another look?

python/feathub/processors/flink/table_builder/join_utils.py
```diff
@@ -71,13 +71,15 @@ def __init__(
 
     @staticmethod
     def from_table_descriptor_and_field_name(
-        table_descriptor: TableDescriptor, field_name: str
+        table_descriptor: TableDescriptor, field_name: str, set_default_value: bool
```
lindong28, line 74: Could you double check whether the python doc of `valid_time_interval` and `default_value` is correct?

It is not intuitive to understand the meaning of `set_default_value` here. Can you explain its semantics?

Would it be more intuitive to get the config from the input TableDescriptor and use this config to determine the expected behavior?

lindong28 (commented):

Sxnan:
@lindong28 After configuration infra is introduced, I updated the PR. Could you please take another look?

python/feathub/processors/flink/table_builder/tests/test_flink_table_builder_native_flink_sliding_window.py
```diff
@@ -0,0 +1,1040 @@
...
+    SlidingWindowTransform,
+)
+from feathub.processors.flink.flink_table import flink_table_to_pandas
+from feathub.processors.flink.table_builder.tests.sliding_window_test_base import (
+    SlidingWindowTestBase,
+)
+from feathub.table.schema import Schema
+
+
+class FlinkTableBuilderNativeFlinkSlidingWindowTransformTest(SlidingWindowTestBase):
```
lindong28, line 40: Can you add python doc explaining the difference between FlinkTableBuilderNativeFlinkSlidingWindowTransformTest and FlinkTableBuilderSlidingWindowTransformTest?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PostSlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.flink.table.runtime.typeutils.ExternalTypeInfo;
+import org.apache.flink.types.Row;
+import org.apache.flink.util.Collector;
+
+import java.time.Instant;
+
+/**
+ * The KeyedProcessFunction that is used after the Flink SQL Sliding Window operation by Feathub.
+ *
+ * <p>The function output the rows with the duplicated values once if deduplicateOutput is true. And
```
lindong28, line 41: Do we still have `deduplicateOutput`?

python/feathub/processors/flink/table_builder/sliding_window_utils.py
```diff
@@ -199,6 +212,14 @@ def evaluate_sliding_window_transform(
...
         )
 
+    if config.get(ENABLE_EMPTY_WINDOW_OUTPUT_CONFIG):
+        result = _apply_post_sliding_window_process_function(
+            result,
+            window_descriptor,
+            agg_descriptors,
+            config.get(ENABLE_STATIC_WINDOW_OUTPUT_CONFIG),
+        )
+
```
lindong28, line 222: To improve code's readability, could you assert `ENABLE_STATIC_WINDOW_OUTPUT_CONFIG == True` here and explain why we don't need to apply the post process function?

python/feathub/registries/registry.py
```diff
@@ -40,18 +40,22 @@ def __init__(self, registry_type: str, config: Dict) -> None:
...
     def build_features(
-        self, features_list: List[TableDescriptor]
+        self, features_list: List[TableDescriptor], props: Optional[Dict] = None
     ) -> List[TableDescriptor]:
         """
         For each table descriptor in the given list, resolve this descriptor by
         recursively replacing its dependent table and feature names with the
-        corresponding table descriptors and features from the cache or registry.
+        corresponding table descriptors and features from the cache or registry. Then
+        recursively configure the table descriptor and its dependent table with the
```
lindong28, line 49: This statement is probably inaccurate because the dependent table will not be configured with the global properties if it is referenced by the given table descriptor by its name (rather than a table descriptor).

lindong28 (commented):
Thanks for the update! Left some minor comments.

Sxnan:
@lindong28 Thank you for your comment! I updated the PR accordingly. Could you please take another look?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -26,6 +28,46 @@
...
+)
+ENABLE_EMPTY_WINDOW_OUTPUT_DOC = (
+    "If it is True, when the sliding window becomes emtpy, it outputs zero value for "
+    "aggregation function SUM and COUNT, and output None for other aggregation "
+    "function. If it is False, the sliding window doesn't output anything when the "
+    "sliding window becomes empty."
+)
+
+ENABLE_STATIC_WINDOW_OUTPUT_CONFIG = (
+    "sdk.sliding_feature_view.enable_static_window_output"
```
lindong28, line 42: Could we change this config name to `sdk.sliding_feature_view.skip_same_window_output`?

Sxnan:
@lindong28 Thanks for your suggestion! I updated the PR. Could you please take another look?

python/feathub/processors/flink/table_builder/tests/sliding_window_test_base.py
```diff
@@ -0,0 +1,45 @@
...
+#  limitations under the License.
+import datetime
+from typing import Dict
+
+from feathub.processors.flink.table_builder.tests.table_builder_test_base import (
+    FlinkTableBuilderTestBase,
+)
+
+
+class SlidingWindowTestBase(FlinkTableBuilderTestBase):
```
lindong28, line 22: Since there is only one subclass of `SlidingWindowTestBase`, do we still need this class?

python/feathub/processors/flink/table_builder/tests/sliding_window_test_base.py
```diff
@@ -0,0 +1,45 @@
...
+    def setUp(self):
+        super().setUp()
+        self.props = self.get_props()
+
+    @staticmethod
+    def _to_epoch_millis(
```
lindong28, line 28: These two static methods seem pretty general.

Would it be useful to move these methods to some util classes so that they can be used by other tests in the future?

lindong28 (commented):
Thanks for the update! LGTM overall. Left just two minor comments.

Sxnan:

lindong28:
Thanks for the update. LGTM.

PR #25, opened 2022-10-31
Update FlinkProcessor cli quickstart

Sxnan:

This PR makes a few updates and fixes a bug to provide a smoother user experience to follow the FlinkProcessor CLI quickstart.

## Brief change log

- Fix the bug by setting serializer of the accumulator of AbstractTimeWindowedAggFunc
- Update the FlinkProcessor CLI quickstart instruction to explicitly lists the prerequisites
- Update the script to build the dependencies to run in CLI mode to avoid using Docker
- Import Kubernetes submitter only in Kubernetes mode
- Specify the version of apache-flink in the setup.py

  - The public API: no

## Documentation


Sxnan:

docs/quickstarts/flink_processor_cli_quickstart.md
```diff
@@ -5,6 +5,13 @@ Commandline Tool to a standalone Flink cluster. The Feathub job simply consumes
 the data from the Flink datagen connector, computes some feature, and prints out the 
 result.
 
+### Prerequisites
+
+In addition to the prerequisites listed in [README](../../README.md#Prerequisites). The
+extra prerequisites to run the quickstart are listed below.
+
+- zip
```
lindong28, line 13: Is this section completed?

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/AbstractTimeWindowedAggFunc.java
```diff
@@ -123,20 +126,28 @@ public TypeInference getTypeInference(DataTypeFactory typeFactory) {
...
+                                                            .get(1)
+                                                            .getLogicalType()));
+                            return Optional.of(
+                                    DataTypes.STRUCTURED(
+                                            TimeWindowedAccumulator.class,
+                                            DataTypes.FIELD(
+                                                    "values",
+                                                    MapView.newMapViewDataType(
+                                                            DataTypes.BIGINT(),
+                                                            DataTypes.RAW(
```
lindong28, line 145: Could you explain the bug? Is it possible to add a unit test for this bug?

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/AbstractTimeWindowedAggFunc.java
```diff
@@ -16,10 +16,13 @@
 
 package com.alibaba.feathub.udf;
 
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.common.typeutils.base.ListSerializer;
 import org.apache.flink.table.api.DataTypes;
 import org.apache.flink.table.api.dataview.MapView;
 import org.apache.flink.table.catalog.DataTypeFactory;
 import org.apache.flink.table.functions.AggregateFunction;
+import org.apache.flink.table.runtime.typeutils.InternalSerializers;
```
lindong28, line 25: This class is @Internal. Is it possible to avoid using internal classes from Flink?

If we have to use an internal class in Flink, it probably suggests Flink's public API lacks the necessary functionalities for users to do the work. Can you describe what is missing here so that we can create a TODO for Flink?

tools/cli-deps/build-cli-deps.sh
```diff
@@ -21,19 +21,11 @@ set -e
...
-mkdir "${CURRENT_DIR}"/wheels
-cp dist/feathub-*.whl "${CURRENT_DIR}"/wheels
-
-docker run -it --rm -v "${CURRENT_DIR}":/build -w /build quay.io/pypa/manylinux2014_x86_64 ./build.sh
+# zip Feathub and its dependencies
+cd "${CURRENT_DIR}"
+# TODO: Install the latest sable version after Feathub released.
+python -m pip install --target __pypackages__ feathub-nightly --no-deps
+python -m pip install --target __pypackages__ -r requirements.txt
+cd __pypackages__ && zip -r deps.zip . && mv deps.zip ../ && cd ..
```
lindong28, line 30: Open-source Flink users who are not using VVP can also submit FeatHub program to Flink clusters using Flink CLI, right?

Would it still be useful to keep the infra for users to prepare zip package using docker, similar to how VVP provides both options to users?

lindong28 (commented):

tools/cli-deps/build-cli-deps.sh
Sxnan: Yes, open source can also submit Feathub programs to Flink cluster using Flink CLI.

To package the Feathub and its dependencies, docker is not required. And it is only required when the dependencies to be installed require some operating systems. I added docs in the CLI quickstart.

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/AbstractTimeWindowedAggFunc.java
Sxnan: I updated the implementation to avoid using internal classes.

Sxnan:
@lindong28 Thank you for your comment. I updated the PR. Could you take another look?

docs/quickstarts/flink_processor_cli_quickstart.md
lindong28: Specifying a command like `zip` in the prerequisite section does not seem common. Is zip installed by default on Unix like operating system such as Linux and Mac OS? 

I am wondering what is the best practice. Is there any existing project that does this in its prerequisite?

docs/quickstarts/flink_processor_cli_quickstart.md
```diff
@@ -5,6 +5,13 @@ Commandline Tool to a standalone Flink cluster. The Feathub job simply consumes
 the data from the Flink datagen connector, computes some feature, and prints out the 
 result.
 
+### Prerequisites
+
+In addition to the prerequisites listed in [README](../../README.md#Prerequisites). The
```
lindong28, line 10: The prerequisites referenced here are needed for building python packages. Do we really need all those prerequisites in order to submit FeatHub job via the Flink CLI tool?

For example, is `Maven` needed to use the Flink CLI tool?

java/feathub-udf/src/main/java/com/alibaba/feathub/udf/AbstractTimeWindowedAggFunc.java
```diff
@@ -132,9 +132,12 @@ public TypeInference getTypeInference(DataTypeFactory typeFactory) {
                                                         MapView.newMapViewDataType(
                                                                 DataTypes.BIGINT(),
                                                                 DataTypes.ARRAY(
-                                                                        callContext
-                                                                                .getArgumentDataTypes()
-                                                                                .get(1)))),
+                                                                                getResultDataType(
```
lindong28, line 135: Should we add a TODO for unit tests according to the offline discussion?

java/feathub-udf/pom.xml
```diff
@@ -34,6 +34,12 @@
             <version>${flink.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
lindong28, line 37: Can we avoid making unnecessary movement changes to the pom file?

lindong28 (commented):

Can you try to improve the PR description so that it is easier to understand the purpose of this PR?

docs/quickstarts/flink_processor_cli_quickstart.md
Sxnan: zip command is indeed trivial and I can find other projects that add this as a prerequisite, even though it may not be installed by default on some Unix like systems, so I remove it from the prerequisite list.

Sxnan:
@lindong28 Thank you for your comment. I updated the PR and the description. Could you take another look?

lindong28:
Thanks for the update. LGTM.

PR #27, opened 2022-11-01
SlidingFeatureView supports setting timestamp field and format

Sxnan:

Currently, SlidingFeatureView uses the timestamp field and format from its source as the timestamp of the output. This unnecessarily tight the sliding window timestamp precision to the source timestamp precision. For example, a source with a timestamp that is up to second percision cannot get the sliding window timestamp with millisecond precision.

This PR updates the SlidingFeatureView so that it supports setting the timestamp field and format of the SlidingFeatureView.

## Brief change log

- Update the `to_java_date_format` utils method to support datetime format with fraction of second
- Update SlidingFeatureView to supports setting timestamp field and format

  - The public API: yes

## Documentation


Sxnan:

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -49,6 +49,8 @@ def __init__(
         name: str,
         source: Union[str, TableDescriptor],
         features: Sequence[Union[str, Feature]],
+        sliding_window_timestamp_field: Optional[str] = None,
```
lindong28, line 52: Since the timestamp generated by sliding window likely have different semantics from the timestamp field in the source table, should we give it a new name instead of re-using source table field name by default?

Since we actually generate a new field in the output, should we also document this field (e.g. its semantic meaning) similar to Flink?

lindong28 (commented):

Sxnan:
@lindong28 Thank you for your comment. I updated the PR. Could you take another look?

python/feathub/feature_views/feature_view.py
```diff
@@ -44,6 +46,12 @@ def __init__(
                          which refers to a feature in the source table.
         :param keep_source_fields: True iff all fields in the source table should be
                                    included in this table.
+        :param timestamp_field: Optional. Optional. If it is not None, it is the
```
lindong28, line 49: There are a couple of issues in this doc. Can you improve it?

Same for `timestamp_format`.

python/feathub/feature_views/feature_view.py
```diff
@@ -56,15 +64,19 @@ def __init__(
             # Uses table's keys as features' keys if features' keys are not specified.
             for feature in [f for f in self.get_resolved_features() if f.keys is None]:
                 feature.keys = keys
+
+        if is_unresolved:
```
lindong28, line 68: Should we merge the `if` statement here with the `if` statement above?

python/feathub/feature_views/feature_view.py
```diff
@@ -56,15 +64,19 @@ def __init__(
...
+            timestamp_field = None
```
lindong28, line 69: What is the purpose of passing `timestamp_field` as inputs to the constructor given that we will anyway overwrite its value here?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -66,12 +69,21 @@ def __init__(
                          or its expression must only depend on the
                          SlidingWindowTransform features specified earlier in this list
                          and their group-by keys.
+        :param sliding_window_timestamp_field: It is the name of the timestamp field of
+                                               the sliding window timestamp. Default to
```
lindong28, line 73: It appears inconsistent with existing docs (and also redundant) to specify the default value here.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -66,12 +69,21 @@ def __init__(
...
+                                               "window_time".
+        :param sliding_window_timestamp_format: It is the format of the sliding window
```
lindong28, line 75: It appears inconsistent to start the doc with `It is` here.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -110,6 +122,8 @@ def to_json(self) -> Dict:
                 feature if isinstance(feature, str) else feature.to_json()
                 for feature in self.features
             ],
+            "sliding_window_timestamp_field": self.timestamp_field,
```
lindong28, line 125: There is no longer any field named `sliding_window_timestamp_field` in this class.

Sxnan:
@lindong28 Thanks for your comments. I updated the PR accordingly. Could you please take another look?

python/feathub/feature_views/feature_view.py
```diff
@@ -44,6 +46,10 @@ def __init__(
                          which refers to a feature in the source table.
         :param keep_source_fields: True iff all fields in the source table should be
                                    included in this table.
+        :param timestamp_field: Optional. If it is not None, it is the name of the field
```
lindong28, line 49: The semantics of the `timestamp_field` in `TableDescriptor` is different from the `timestamp_field` here. For `TableDescriptor`, if it is None, then the TableDescriptor won't have any field that represents when the row is generated. For `FeatureView`, if it is None, then we will still try to re-use the `timestamp_field` from the source table.

Thus we can not simply re-use the doc from `TableDescriptor`.

How about this:

Optional. If not None, the feature with the given name is used as the timestamp_field of the TableDescriptor represented by this FeatureView. Otherwise, the timestamp_field of the source TableDescriptor is used as the timestamp_field of the TableDescriptor represented by this FeatureView.

python/feathub/feature_views/feature_view.py
```diff
@@ -44,6 +46,10 @@ def __init__(
...
+                                whose values show the time when the corresponding row
+                                is generated.
+        :param timestamp_format: The format of the timestamp field.
```
lindong28, line 52: This doc seems confusing. If a FeatureView is instantiated with `timestamp_field=None` and `timestamp_format=""%Y-%m-%d %H:%M:%S.%f""`, what is the expected format of the table descriptor represented by this FeatureView?

python/feathub/feature_views/feature_view.py
```diff
@@ -56,15 +62,16 @@ def __init__(
             # Uses table's keys as features' keys if features' keys are not specified.
             for feature in [f for f in self.get_resolved_features() if f.keys is None]:
                 feature.keys = keys
+
+            if timestamp_field is None:
+                timestamp_field = cast(TableDescriptor, self.source).timestamp_field
```
lindong28, line 67: What happens when this feature view is un-resolved and the source table descriptor is a string? Will it raise an error or simply assign None to the timestamp_field?

Can you add a unit test if there is a bug here?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -66,12 +69,17 @@ def __init__(
                          or its expression must only depend on the
                          SlidingWindowTransform features specified earlier in this list
                          and their group-by keys.
+        :param timestamp_field: The name of the field whose values show the time when
```
lindong28, line 72: This doc seems to suggest that this string refers to an existing feature name in the `source` or `features`, which is misleading. This is unlike the timestamp_field of other classes which simply refers to an existing feature name in their inputs.

How about this: 

The name of the field generated by the SlidingFeatureView that represents the closed window end time for each row.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -66,12 +69,17 @@ def __init__(
...
+                                the corresponding sliding window row is generated.
+        :param timestamp_format: The format of the timestamp field.
         """
         super().__init__(
             name=name,
             source=source,
             features=features,
```
lindong28, line 79: Should we add to the `features` a new feature that represents the window end time generated by the SlidingFeatureView?

Otherwise, if we simply pass this timestamp field to FeatureView without adding a feature, we end up with a TableDescriptor whose `self.timestamp_field` can not be found in its `self.get_output_fields(...)`.

If this is indeed a bug, can you add a unit test to cover it in the latest PR?

lindong28 (commented):

python/feathub/feature_views/sliding_feature_view.py
Sxnan: Good catch! It is indeed a bug. I add the window_time feature to the init method of SlidingFeatureView. Unit test is also added to cover the case.

python/feathub/feature_views/feature_view.py
Sxnan: It will simply assign None to the `timestamp_field`. I added a test to ensure the behavior.

Sxnan:
@lindong28 Thanks for your comment. I updated the PR accordingly. Could you take another look?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -40,15 +45,18 @@ class SlidingFeatureView(FeatureView):
...
+    field with default name `window_time` and format "%Y-%m-%d %H:%M:%S.%f". Other
+    fields in the source table will not be included.
     """
 
     def __init__(
         self,
         name: str,
         source: Union[str, TableDescriptor],
         features: Sequence[Union[str, Feature]],
+        timestamp_field: str = "window_time",
```
lindong28, line 58: Would it be better to use `window_end` to indicate that this is the time when the window end?

Note that Flink's `Hop` table function outputs the `window_start` and `window_end` columns.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -40,15 +45,18 @@ class SlidingFeatureView(FeatureView):
...
+        timestamp_format: str = "%Y-%m-%d %H:%M:%S.%f",
```
lindong28, line 59: More existing classes (e.g. `TableDescriptor`) use `epoch` as the default value for `timestamp`. Would it be more consistent and more performant to use the same default value here?

python/feathub/processors/flink/table_builder/tests/test_flink_table_builder_sliding_window.py
```diff
@@ -581,21 +582,106 @@ def test_transform_with_value_counts(self):
...
+                        limit=3,
+                    ),
+                ),
+                Feature(
+                    name="epoch_window_time",
+                    dtype=Int64,
+                    transform="UNIX_TIMESTAMP(sliding_window_timestamp)",
+                ),
+            ],
+            timestamp_field="sliding_window_timestamp",
```
lindong28, line 642: Can you add a unit test with SlidingWindow's `timestamp_format` different from the timestamp format of its source field?

python/feathub/feature_views/transforms/feature_view_generated_feature_transform.py
```diff
@@ -0,0 +1,31 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from typing import Dict
+
+from feathub.feature_views.transforms.transformation import Transformation
+
+
+class _FeatureViewGeneratedFeatureTransform(Transformation):
```
lindong28, line 19: Instead of adding a new subclass of `Transform`, would it be simpler to just reuse `ExpressionTransform` and add a built-in-function that generates the timestamp value?

python/feathub/feature_views/sliding_feature_view.py
Sxnan: `epoch` is not the default value of `timestamp_format`. A TODO is added to support epoch with millisecond precision.

python/feathub/feature_views/transforms/feature_view_generated_feature_transform.py
Sxnan: I think it makes sense to add a built-in function such as `CURRENT_EVENT_TIME` that generates the timestamp value. 

However, I found that it is not trivial to support it with the current implementation of FlinkAstEvaluator, which naively passes the function call to Flink SQL expression. And the `CURRENT_EVENT_TIME` doesn't exist in Flink SQL expression. We need to refactor FlinkAstEvaluator to properly handle the `CURRENT_EVENT_TIME` function, which requires non-trivial work. And I think it should not block this PR. 

Therefore, the `CURRENT_EVENT_TIME` is only used internally for now. And we can expose it as a built-in function to users after we can handle `CURRENT_EVENT_TIME` properly. What do you think?

Sxnan:
@lindong28 Thanks for your comment. I have updated the PR. Could you please take another look?

python/feathub/feature_views/transforms/feature_view_generated_feature_transform.py
lindong28: Sounds good. Thanks for the explanation.

lindong28:
Thanks for the update! LGTM.

PR #31, opened 2022-11-07
Support ProcessorJob cancel

Sxnan:

This PR adds a cancel method to ProcessorJob to support canceling an unbounded stream job.

## Brief change log

- Add abstract "cancel" method to `ProcessorJob`
- Implement cancel method for `LocalProcessor` and `FlinkProcessor`
- Refactor FlinkProecssorJobs to the corresponding submitters

  - The public API: yes
## Documentation


Sxnan:

python/feathub/processors/processor_job.py
```diff
@@ -24,6 +24,16 @@ class ProcessorJob(ABC):
     def __init__(self) -> None:
         pass
 
+    @abstractmethod
+    def cancel(self) -> None:
```
lindong28, line 28: Would it useful to return a `Future` similar to `pyflink.common.JobClient.cancel()`?

lindong28 (commented):
Thanks for the PR.

python/feathub/processors/processor_job.py
Sxnan: After looking at the `pyflink.common.JobClient`, I think the Future is only used for acknowledgment. It is completed as soon as the server confirms the cancel is received but not when the job is canceled. 

Currently, the cancel method implies the same semantics as the method should return after the cancel is acknowledged. We may need to state that in the docstring explicitly. Given that the two ways have the same semantics and we are currently not using `CompletableFuture` in other places, I am leaning toward the current way. What do you think?

Sxnan: After the offline discussion, I agree it is better to return the Future object and users can decide if they want to wait for acknowledgment or not.

Sxnan:
@lindong28 Thank you for your review! I have updated the PR. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #34, opened 2022-11-11
KafkaSource supports bounded mode

Sxnan:

This PR aims to solve the problem that the Table#to_pandas method never returns when using KafkaSource by supporting bounded mode for Flink Kafka. User can now use bounded KafkaSource so that Table#to_pandas method can return.

Currently, the native Flink Kafka SQL connector doesn't support bounded mode. There is a [ticket](https://issues.apache.org/jira/browse/FLINK-24456) tracking the issue. As a temporary solution, we introduce a `BoundedKafkaDynamicSource` that extends the native `KafkaDynamicSource` to support bounded mode.

## Brief change log

- Introduce `BoundedKafkaDynamicSource`
- Use testcontainers for Kafka test

  - The public API: no

## Documentation

### DIFF
--- a/java/feathub-connectors/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/BoundedKafkaDynamicTableFactory.java
+++ b/java/feathub-connectors/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/BoundedKafkaDynamicTableFactory.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2022 The Feathub Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.flink.streaming.connectors.kafka.table;
+
+import org.apache.flink.api.common.serialization.DeserializationSchema;
+import org.apache.flink.streaming.connectors.kafka.config.StartupMode;
+import org.apache.flink.streaming.connectors.kafka.internals.KafkaTopicPartition;
+import org.apache.flink.table.connector.format.DecodingFormat;
+import org.apache.flink.table.data.RowData;
+import org.apache.flink.table.types.DataType;
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.regex.Pattern;
+
+// TODO: This class will be removed after FLINK-24456 resolved.
+/**
+ * Factory for creating configured instances of {@link BoundedKafkaDynamicSource} and {@link
+ * KafkaDynamicSink}.
+ */
+public class BoundedKafkaDynamicTableFactory extends KafkaDynamicTableFactory {
+
+    public static final String IDENTIFIER = "bounded-kafka";
+
+    @Override
+    public String factoryIdentifier() {
+        return IDENTIFIER;
+    }
+
+    @Override
+    protected BoundedKafkaDynamicSource createKafkaTableSource(
+            DataType physicalDataType,
+            @Nullable DecodingFormat<DeserializationSchema<RowData>> keyDecodingFormat,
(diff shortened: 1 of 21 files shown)


Sxnan:

python/feathub/processors/flink/table_builder/tests/test_kafka_utils.py
```diff
@@ -192,19 +213,97 @@ def test_kafka_source_sink(self):
...
+        features = DerivedFeatureView(
+            "feature_view", source, features=[], keep_source_fields=True
+        )
+        table = table_builder.build(features, bounded_source=False)
+        table_result = table.execute()
+        with table_result.collect() as results:
+            result_rows = set(results)
+        self.assertEquals(expected_rows, result_rows)
+
+    def _prepare_data_to_kafka(self, t_env):
```
lindong28, line 286: nits: it might be more readable to name it `_produce_data_to_kafka()`.

python/feathub/processors/flink/table_builder/file_system_utils.py
```diff
@@ -67,6 +67,7 @@ def get_table_from_file_source(
         # Set ignore-parse-errors to set null in case of csv parse error
         descriptor_builder.option("csv.ignore-parse-errors", "true")
 
+    t_env.drop_temporary_table(file_source.name)
```
lindong28, line 70: Why do we need to drop the table here?

python/feathub/processors/flink/table_builder/flink_table_builder.py
```diff
@@ -103,6 +103,7 @@ def build(
         keys: Union[pd.DataFrame, TableDescriptor, None] = None,
         start_datetime: Optional[datetime] = None,
         end_datetime: Optional[datetime] = None,
+        bounded_source: bool = False,
```
lindong28, line 106: Should we be able to get this information from `features`, which in turn can get this information from e.g. `KafkaSource`?

lindong28 (commented):
Thanks for the PR! Looks good overall. Left some comments below.

python/feathub/processors/flink/table_builder/file_system_utils.py
Sxnan: StreamTableEnvironment throws an exception when creating temporary tables with the same name so the temporary table is dropped before we create a new one.

After some thought, I think it is better to append a randomly generated part to the temporary table name to avoid name conflict, similar to what we did for Sink FeatureTable. Similar to sink, we will use an anonymous table after the bug in VVR-6.0.2 is fixed so that we can get rid of the `create_temporary_table` altogether.

Sxnan:

Sxnan:
@lindong28 I removed the change to Table#to_pandas and updated the PR title and description. Could you please take another look?

python/feathub/processors/flink/table_builder/source_sink_utils_common.py
```diff
@@ -27,8 +27,8 @@
 )
 
 
-def generate_random_table_name() -> str:
-    random_sink_name = "anonymous" + str(uuid.uuid4()).replace("-", "") + "table"
+def generate_random_table_name(table_id: str) -> str:
```
lindong28, line 30: How about changing the parameter to `name_prefix`?

java/feathub-connectors/flink-connectors/flink-connector-kafka/src/main/java/org/apache/flink/streaming/connectors/kafka/table/BoundedKafkaDynamicTableFactory.java
```diff
@@ -0,0 +1,77 @@
...
+
+import javax.annotation.Nullable;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Properties;
+import java.util.regex.Pattern;
+
+/**
+ * Factory for creating configured instances of {@link BoundedKafkaDynamicSource} and {@link
```
lindong28, line 34: Can we add a TODO to remove this class after FLINK-24456 is resolved?

java/feathub-connectors/flink-connectors/flink-connector-kafka/src/test/java/org/apache/flink/streaming/connectors/kafka/table/KafkaTableTestBase.java
```diff
@@ -0,0 +1,117 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+import java.time.Duration;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Properties;
+
+/**
+ * Base class for Kafka Table IT Cases.
```
lindong28, line 43: Can we add a TODO to remove this class after FLINK-24456 is resolved?

lindong28 (commented):

Sxnan:
@lindong28 Thanks for your suggestions! I updated the PR. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #38, opened 2022-11-17
Table#to_pandas supports force bounded

Sxnan:

Currently, Table#to_pandas never returns if the Table is unbounded. 

In this PR, we make the `to_pandas` method raise an exception when the Table is unbounded. And we add a `force_bounded` argument to the method to allow users to force the unbounded table to be bounded and convert to Pandas DataFrame. 

## Brief change log

- Add boundedness API to TableDescriptor
- Table#to_pandas supports force bounded

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/feature_tables/feature_table.py
+++ b/python/feathub/feature_tables/feature_table.py
@@ -16,8 +16,8 @@ from typing import List, Optional, Dict, Any
 
 from feathub.common.exceptions import FeathubException
 from feathub.feature_views.feature import Feature
-from feathub.table.table_descriptor import TableDescriptor
 from feathub.table.schema import Schema
+from feathub.table.table_descriptor import TableDescriptor
 
 
 # TODO: Update Sink implementation of FeatureTable to support rename timestamp field
@@ -92,4 +92,16 @@ class FeatureTable(TableDescriptor, ABC):
             dtype=self.schema.get_field_type(feature_name),
             transform=feature_name,
             keys=self.keys,
+        )
+
+    def is_bounded(self) -> bool:
+        # FeatureTable is bounded by default unless subclass overwrite the method.
+        return True
+
+    def get_bounded_view(self) -> TableDescriptor:
+        if self.is_bounded():
+            return self
+
+        raise FeathubException(
+            f"{type(self)} is unbounded and it doesn't support getting bounded view."
         )
--- a/python/feathub/feature_tables/sinks/kafka_sink.py
+++ b/python/feathub/feature_tables/sinks/kafka_sink.py
@@ -13,10 +13,10 @@
 #  limitations under the License.
 from typing import Dict, Optional
 
-from feathub.feature_tables.feature_table import FeatureTable
+from feathub.feature_tables.sinks.sink import Sink
 
 
-class KafkaSink(FeatureTable):
+class KafkaSink(Sink):
     def __init__(
         self,
         bootstrap_server: str,
--- a/python/feathub/feature_tables/sources/datagen_source.py
+++ b/python/feathub/feature_tables/sources/datagen_source.py
@@ -11,12 +11,16 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from copy import deepcopy
 from datetime import timedelta
 from typing import Dict, Optional, List, Union, Any
 
 from feathub.common.exceptions import FeathubException
 from feathub.feature_tables.feature_table import FeatureTable
 from feathub.table.schema import Schema
+from feathub.table.table_descriptor import TableDescriptor
+
+DEFAULT_BOUNDED_NUMBER_OF_ROWS = 100
 
 
 class RandomField:
@@ -75,53 +79,18 @@ class SequenceField:
         return {"type": "sequence", "start": self.start, "end": self.end}
 
 
-default_field_config = RandomField()
-
-
-class DataGenConfig:
-    """
-    DataGenConfig specifies how the data are generated.
-    """
-
-    def __init__(
-        self,
-        rows_per_second: int = 10000,
-        number_of_rows: Optional[int] = None,
-        field_configs: Optional[Dict[str, Union[RandomField, SequenceField]]] = None,
-    ) -> None:
-        """
-        :param rows_per_second: Rows per second to control the emit rate.
-        :param number_of_rows: Optional. If it is None, unlimited number of rows will be
-                               generated. If it is not None, it specifies the total
-                               number of rows to emit.
-        :param field_configs: A Map of field to the config of the field. The config
-                              can be either RandomField or SequenceField. Every field
-                              should be in the schema of the DataGenSource. If a field
-                              in the schema doesn't have a config, it is set to
-                              `default_field_config`.
-        """
-        self.rows_per_second = rows_per_second
-        self.number_of_rows = number_of_rows
-        self.field_configs = field_configs
-
-    def to_json(self) -> Dict:
-        return {
-            "rows_per_second": self.rows_per_second,
-            "number_of_rows": self.number_of_rows,
-            "field_configs": {k: v.to_json() for k, v in self.field_configs.items()},
-        }
-
-
 class DataGenSource(FeatureTable):
     """
-    DataGenSource generate table with random data or sequential data.
+    DataGenSource generates table with random data or sequential data.
     """
 
     def __init__(
         self,
         name: str,
         schema: Schema,
-        data_gen_config: DataGenConfig = DataGenConfig(),
+        rows_per_second: int = 10000,
+        number_of_rows: Optional[int] = None,
+        field_configs: Optional[Dict[str, Union[RandomField, SequenceField]]] = None,
         keys: Optional[List[str]] = None,
         timestamp_field: Optional[str] = None,
         timestamp_format: str = "epoch",
@@ -130,8 +99,15 @@ class DataGenSource(FeatureTable):
         """
         :param name: The name that uniquely identifies this source in a registry.
         :param schema: The schema of the data.
-        :param data_gen_config: The DataGenConfig that specify how the data are
-                                generated.
+        :param rows_per_second: Rows per second to control the emit rate.
+        :param number_of_rows: Optional. If it is None, unlimited number of rows will be
+                               generated. If it is not None, it specifies the total
+                               number of rows to emit.
+        :param field_configs: A Map of field to the config of the field. The config
+                              can be either RandomField or SequenceField. Every field
+                              should be in the schema of the DataGenSource. If a field
+                              in the schema doesn't have a config, it is set to
+                              RandomField.
         :param keys: Optional. The names of fields in this feature view that are
                      necessary to interpret a row of this table. If it is not None, it
                      must be a superset of keys of any feature in this table.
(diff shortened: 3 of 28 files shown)


Sxnan:

python/feathub/feature_tables/sources/datagen_source.py
```diff
@@ -102,7 +105,22 @@ def __init__(
         """
         self.rows_per_second = rows_per_second
         self.number_of_rows = number_of_rows
-        self.field_configs = field_configs
+        self.field_configs = field_configs if field_configs is not None else {}
+
+    def check_bounded(self) -> bool:
```
lindong28, line 110: How about name this method as `is_bounded()`?

python/feathub/feature_tables/sources/file_system_source.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         timestamp_field: Optional[str] = None,
         timestamp_format: str = "epoch",
         max_out_of_orderness: timedelta = timedelta(0),
+        is_bounded: bool = True,
```
lindong28, line 35: Is there any case where we should initialize a FileSystemSource with `is_bounded=False`?

python/feathub/feature_tables/sources/online_store_source.py
```diff
@@ -39,13 +40,17 @@ def __init__(
...
+            is_bounded=True,
             keys=keys,
             timestamp_field=None,
             timestamp_format="epoch",
         )
         self.store_type = store_type
         self.table_name = table_name
 
+    def get_bounded_table(self) -> TableDescriptor:
+        return OnlineStoreSource(self.name, self.keys, self.store_type, self.table_name)
```
lindong28, line 52: Return `self` here?

python/feathub/feature_views/transforms/join_transform.py
```diff
@@ -22,21 +22,23 @@ class JoinTransform(Transformation):
     """
 
     def __init__(
-        self,
-        table_name: str,
-        feature_name: str,
+        self, table_name: str, feature_name: str, force_table_bounded: bool = False
```
lindong28, line 25: How about we rename `force_table_bounded` as `is_bounded`?

python/feathub/table/table_descriptor.py
```diff
@@ -79,3 +79,21 @@ def get_feature(self, feature_name: str) -> Feature:
...
+        """
+        Gets a copy of self that is bounded. All sources of the return TableDescriptor
+        are bounded.
+
+        :return: A bounded table descriptor.
+        """
+        pass
+
+    @abstractmethod
+    def check_bounded(self, registry: Registry) -> bool:
```
lindong28, line 94: How about we name this method as `is_bounded()`?

python/feathub/feature_views/feature_view.py
```diff
@@ -169,3 +172,30 @@ def _get_keys(self) -> Optional[List[str]]:
...
+        return all(
+            [
+                cast(TableDescriptor, self.source).check_bounded(registry),
+                *[
+                    feature.transform.force_table_bounded
+                    or registry.get_features(
+                        feature.transform.table_name
+                    ).check_bounded(registry)
+                    for feature in self.get_resolved_features()
+                    if isinstance(feature.transform, JoinTransform)
```
lindong28, line 186: For join transform, suppose source is bounded and dimension table is unbounded (e.g. Kafka Stream), the output should still be bounded, right?

If yes, should we remove `force_table_bounded` from `JoinTransform`?

python/feathub/feature_views/on_demand_feature_view.py
```diff
@@ -43,11 +43,15 @@ def __init__(self) -> None:
...
                 system_name="online_request",
                 properties={},
+                is_bounded=True,
                 keys=None,
                 timestamp_field=None,
                 timestamp_format="epoch",
             )
 
+        def get_bounded_table(self) -> TableDescriptor:
+            return OnDemandFeatureView._OnlineRequestSource()
```
lindong28, line 53: Return `self`?

python/feathub/processors/flink/flink_table.py
```diff
@@ -118,17 +118,24 @@ def __init__(
...
-        return flink_table_to_pandas(self._flink_table)
+        if not self.feature.check_bounded(self.flink_processor.registry):
+            if not force_bounded:
+                raise FeathubException(
+                    "Unbounded table cannot be converted to Pandas DataFrame. You can "
+                    "set force_bounded to True to convert the Table to DataFrame."
+                )
+
+        return flink_table_to_pandas(
+            self._get_flink_table(self.feature.get_bounded_table())
```
lindong28, line 137: Should we call `get_bounded_table()` only when `force_bounded = True`?

lindong28 (commented):

Sxnan:

python/feathub/table/table_descriptor.py
```diff
@@ -79,3 +79,21 @@ def get_feature(self, feature_name: str) -> Feature:
...
+        """
+        Gets a copy of self that is bounded. All sources of the return TableDescriptor
+        are bounded.
+
+        :return: A bounded table descriptor.
+        """
+        pass
+
+    @abstractmethod
+    def is_bounded(self, registry: Registry) -> bool:
```
lindong28, line 94: This method requires `registry` as input because `FeatureView` needs it to get around the problem described in FLINK-30078. 

Can we avoid complicating the public API due to a bug in Flink? For example, maybe we can just return `source.is_bounded()` here, and raise exception in `feature_view.build()` as appropriate.

We can also remove `_force_table_bounded` from e.g. `JoinTransform` by doing so.

python/feathub/feature_tables/sources/file_system_source.py
```diff
@@ -54,16 +55,20 @@ def __init__(
...
-            schema=schema,
             timestamp_field=timestamp_field,
             timestamp_format=timestamp_format,
+            schema=schema,
         )
         self.path = path
         self.schema = schema
         self.max_out_of_orderness = max_out_of_orderness
 
+    def get_bounded_view(self) -> TableDescriptor:
```
lindong28, line 69: Instead of having every bounded source duplicate this method, would it be simpler to just add a default implementation in the parent class with the following logic:

Return self if is_bounded() == True. Otherwise raise exception.

python/feathub/feature_tables/feature_table.py
```diff
@@ -73,6 +79,7 @@ def __init__(
         self.system_name = system_name
         self.properties = properties
         self.schema = schema
+        self._is_bounded = is_bounded
```
lindong28, line 82: Since `is_bounded` is a feature that can be derived from existing fields and the actual class type, would it be simpler to keep `FeatureTable` compact without adding this field? We keep the `is_bounded` only for those classes that actually need it (e.g. KafkaSource).

We would also skip passing `is_bounded=True` for many classes.

python/feathub/feature_tables/sources/kafka_source.py
```diff
@@ -115,7 +117,26 @@ def __init__(
             raise FeathubException(
                 "startup_datetime is required when startup_mode is timestamp."
             )
-        self.is_bounded = is_bounded
+
+    def get_bounded_view(self) -> TableDescriptor:
+        return KafkaSource(
```
lindong28, line 122: Can we use `deepcopy(self)` similar to what this PR do in `FeatureView::get_bounded_view(...)`?

lindong28 (commented):

python/feathub/feature_tables/sources/datagen_source.py
```diff
@@ -121,7 +139,7 @@ def __init__(
         self,
         name: str,
         schema: Schema,
-        data_gen_config: DataGenConfig = DataGenConfig(),
+        data_gen_config: Optional[DataGenConfig] = None,
```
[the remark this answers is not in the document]
Sxnan: You are right that it is simpler to remove `DataGenSourceConfig`.

Sxnan:
@lindong28 Thanks for your comment! I updated the PR. Could you please take another look?

python/feathub/feature_tables/sinks/kafka_sink.py
```diff
@@ -53,6 +53,9 @@ def __init__(
             {} if producer_properties is None else producer_properties
         )
 
+    def is_bounded(self) -> bool:
```
lindong28, line 56: After thinking about this more, I find it counter-intuitive to mark a sink as bounded or unbounded. This is because boundedness is only meaningful is the object is readable. But a sink is typically not readable, as it typically won't have the attributes (e.g. kafka consumer properties) needed when consuming the data.

How about we just overwrite this method in sink.py such that this method throws exception for all sinks, similar to Sink::get_bounded_view()?

python/feathub/feature_tables/sources/datagen_source.py
```diff
@@ -75,21 +79,26 @@ def to_json(self) -> Dict:
...
 
 
-default_field_config = RandomField()
-
-
-class DataGenConfig:
+class DataGenSource(FeatureTable):
     """
-    DataGenConfig specifies how the data are generated.
+    DataGenSource generate table with random data or sequential data.
```
lindong28, line 84: nits: generate -> generates

python/feathub/feature_tables/sources/datagen_source.py
```diff
@@ -144,33 +120,55 @@ def __init__(
...
 
     def to_json(self) -> Dict:
         return {
             "type": "DataGenSource",
             "name": self.name,
             "schema": self.schema,
-            "data_gen_config": self.data_gen_config.to_json(),
+            "rows_per_second": self.rows_per_second,
+            "number_of_rows": self.number_of_rows,
+            "field_configs": {k: v.to_json() for k, v in self.field_configs.items()},
```
lindong28, line 171: Should we explicitly handle the case where `field_configs` is `None`?

python/feathub/feature_views/feature_view.py
```diff
@@ -169,3 +170,18 @@ def _get_keys(self) -> Optional[List[str]]:
                 key_fields.extend(keys)
 
         return list(OrderedDict.fromkeys(key_fields))
+
+    def is_bounded(self) -> bool:
+        return cast(TableDescriptor, self.source).is_bounded()
```
lindong28, line 175: Should we re-use `get_resolved_source()` here so that it throws exception if the source is unresolved?

python/feathub/feature_views/feature_view.py
```diff
@@ -169,3 +170,18 @@ def _get_keys(self) -> Optional[List[str]]:
...
+
+    def get_bounded_view(self) -> TableDescriptor:
+        if self.is_bounded():
+            return self
+
+        if self.is_unresolved():
```
lindong28, line 181: It seems simpler to remove this since `self.is_bounded()` should have thrown an exception if the table is unsolved.

python/feathub/processors/flink/table_builder/tests/test_kafka_utils.py
```diff
@@ -101,7 +101,7 @@ def test_kafka_source(self):
                 expected_options, dict(flink_table_descriptor.get_options())
             )
 
-            source.is_bounded = True
+            source._is_bounded = True
```
lindong28, line 104: It is preferred to use FeatHub public API in tests so that the tests mimic how user code interacts with FeatHub.

How about we use `source.get_bounded_view()` here?

python/feathub/table/table_descriptor.py
```diff
@@ -79,3 +79,19 @@ def get_feature(self, feature_name: str) -> Feature:
...
+
+        :return: A bounded table descriptor.
+        """
+        pass
+
+    @abstractmethod
+    def is_bounded(self) -> bool:
+        """
+        Whether the Table is bounded.
+        """
```
lindong28, line 97: Should we add `pass` for consistency with other abstract methods?

python/feathub/table/table.py
```diff
@@ -49,9 +49,11 @@ def get_schema(self) -> Schema:
...
 
     @abstractmethod
-    def to_pandas(self) -> pd.DataFrame:
+    def to_pandas(self, force_bounded: bool = False) -> pd.DataFrame:
         """
         Returns a Pandas DataFrame containing values of this table.
+
+        :param force_bounded: Whether to force the table to be bounded.
         """
         return pd.DataFrame()
```
lindong28, line 58: Can you help change this to `pass` for consistency with other abstract methods?

lindong28 (commented):

python/feathub/feature_tables/sources/datagen_source.py
Sxnan: This self.data_gen_config should not be null. We assign an empty dict in the __init__ method if the user doesn't give one.

Sxnan:
@lindong28 Thanks for your comments! I updated the PR. Could you please take another look?

python/feathub/feature_tables/feature_table.py
```diff
@@ -93,3 +93,15 @@ def get_feature(self, feature_name: str) -> Feature:
             transform=feature_name,
             keys=self.keys,
         )
+
+    def is_bounded(self) -> bool:
+        # FeatureTable is bounded by default unless subclass overwrite the method.
+        return True
+
+    def get_bounded_view(self) -> "FeatureTable":
```
lindong28, line 101: Is there any specific benefit of having this method return `FeatureTable` instead of `TableDescriptor`?

If not, it seems simpler and more consistent with the abstract method definition to just return `TableDescriptor` here?

Same for other `get_bounded_view()` overrides.

lindong28 (commented):
Thanks for the update. LGTM overall. Left one minor comment.

python/feathub/feature_tables/feature_table.py
Sxnan: I think it is common in Java to return covariant type when overriding methods. Although not obvious, the benefit is that the user/developer can get a more specific return type when calling the method in the subclass, which can avoid type casting. [1](https://www.whizlabs.com/blog/covariant-return-types-in-java/)[2](https://www.section.io/engineering-education/method-overriding-and-overloading-using-covariant-return-type-in-java/). What do you think? 

Currently, in unit tests, we can avoid typecasting.

lindong28: Hmm.. can you explain which test may fail (due to lack of typecasting) if we update all `get_bounded_view()` overrides to return `TableDescriptor`?

Covariant type is useful if the subclass has a public method not found in the parent class which we want to invoke after calling `get_bounded_view()`. Do you have a concrete use-case here?

Sxnan: It is line 107 in `test_kafka_utils.py`, where calling the `get_table_from_source` requires a `FeatureTable` but `get_bounded_view` returns `TableDescriptor`. 

I think the subclass has no public method that is not found in the parent class. And due to the lack of a use case here. You are right that we should prefer consistency here. I have updated the return type of the `get_bounded_view()` overrides.

Sxnan:
@lindong28 Thanks you for your suggestion! I have updated the PR accordingly. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #39, opened 2022-11-21
FlinkProcessor support timestamp with time zone

Sxnan:

This PR update FlinkProcessor so that it supports timestamp with time zone. The PR also updates the FlinkProcessor implementation so that it uses TIMESTAMP_LTZ(3) instead of TIMESTAMP(3) as row time attribute type. The TIMESTAMP_LTZ is considered safer to use, as it explicitly defines the timezone, and it supports [daylight-saving time](https://nightlies.apache.org/flink/flink-docs-master/docs/dev/table/timezone/#daylight-saving-time-support).   

## Brief change log

- FlinkProcessor internally use type TIMESTAMP_LTZ(3) instead of TIMESTAMP(3) as row time attribute type
- to_java_date_format supports format with time zone

  - The public API: no

## Documentation


lindong28:
Thanks for the PR! LGTM.

PR #42, opened 2022-11-29
FlinkProcessor source utils method generates random source table name to avoid conflict

Sxnan:

FlinkProcessor source utils method generates a random table name for the temporary source table to avoid conflict.

This bug causes the nightly build to fail lately.

## Brief change log

- Update `datagen_utils.py` and `file_system_utils.py` to generate a random table name for the temporary source table 

  - The public API: no

## Documentation


Sxnan:

python/feathub/processors/flink/table_builder/datagen_utils.py
```diff
@@ -99,6 +100,6 @@ def get_table_from_data_gen_source(
         else:
             raise FeathubException(f"Unknown field config type {type(field_config)}.")
 
-    t_env.create_temporary_table(data_gen_source.name, table_descriptor_builder.build())
-
-    return t_env.from_path(data_gen_source.name)
+    table_name = generate_random_table_name(data_gen_source.name)
```
lindong28, line 103: Would it be useful to update `generate_random_table_name()` so that it is easy for us to programmably derive the original table name from the random table name?
 
For example, it can be in the format {original_table_name}-{uuid}.

Sxnan:
@lindong28 Thanks for your comment! I updated the PR. Could you take another look?

lindong28:
Thanks for the update! LGTM.

PR #45, opened 2022-12-02
Share t_env and env across tests

### LINKED ISSUE #24: FeatHub unit tests should not leak file descriptor
Running `pytest -W ignore::DeprecationWarning` might fail with the "Too many open files" error.

It appears that the number of open files with "PIPE" type keeps increasing over time. This can be verified by running the following command repeatedly after starting pytest.

```
ps aux | grep pytest | grep local | awk '{print $2}' | xargs lsof -p | grep PIPE  | wc -l
```

Sxnan:

This PR mitigate the file descriptor leak when running unit test. This PR fix #24.

## Brief change log

- Share t_env and env across tests

  - The public API: no

## Documentation


lindong28:
Thanks for the PR! LGTM.

PR #47, opened 2022-12-06
Move the join transform boundedness check to FlinkTableBuilder

Sxnan:

Currently, the boundedness of the left table and right table of JoinTransformation is checked when building the DerivedFeatureView.

The current implementation fails to raise an exception in the case of calling `to_pandas(force_bounded=True)` on a built unbounded left table that joins an unbounded right table. That's because, in the `to_pandas` method, we get a bounded feature view from the built unbounded table, and we don't check the bounded feature view and the unbounded right table to join.

More importantly, it exposes a processor-specific limitation to the DerivedFeatureView API.

Therefore, we move the boundedness check to FlinkTableBuilder in this PR.

## Brief change log

- Move the join transform boundedness check to FlinkTableBuilder

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/processors/flink/table_builder/tests/test_flink_table_builder_derived_feature.py
+++ b/python/feathub/processors/flink/table_builder/tests/test_flink_table_builder_derived_feature.py
@@ -16,7 +16,9 @@ from datetime import timedelta
 
 import pandas as pd
 
+from feathub.common.exceptions import FeathubException
 from feathub.common.types import Float64, Int64, String
+from feathub.feature_tables.sources.datagen_source import DataGenSource
 from feathub.feature_views.derived_feature_view import DerivedFeatureView
 from feathub.feature_views.feature import Feature
 from feathub.feature_views.transforms.over_window_transform import OverWindowTransform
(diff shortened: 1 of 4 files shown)


Sxnan:

python/feathub/processors/flink/table_builder/tests/test_flink_table_builder_derived_feature.py
```diff
@@ -512,3 +514,38 @@ def test_join_transform_with_zoned_timestamp(self):
...
+            name="feature_view_1",
+            source=source,
+            features=["source_2.val2"],
+            keep_source_fields=True,
+        )
+
+        built_feature_view = self.registry.build_features([source_2, feature_view_1])[1]
+
+        with self.assertRaises(FeathubException) as cm:
+            self.flink_table_builder.build(built_feature_view.get_bounded_view())
```
lindong28, line 545: Would it be simpler to define `source` as a bounded table instead of calling `get_bounded_view()` here?

Sxnan:
@lindong28 Thank you for your suggestion. I updated the PR and the description. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #50, opened 2022-12-12
SlidingFeatureView supports sliding window with different sizes

Sxnan:

This PR update the SlidingFeatureView and FlinkTableBuilder to support applying sliding window with different sizes in one SlidingFeatureView. 

The implementation applies a pre-aggregation with a tumbling window whose window size is the same as the step size of the sliding window to be applied. Then sliding window only stores the rows for the maximum window size, instead of rows for each sliding window, which uses less space in the Flink state backend than the approach that applies multiple sliding windows and then joins them together.

## Brief change log

- Introduce Java implementation of MultiSizeSlidingWindowKeyedProcessFunction.
- SlidingFeatureView supports sliding window with different sizes

  - The public API: no

## Documentation

### DIFF
--- a/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
+++ b/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2022 The Feathub Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.feathub.flink.udf;
+
+import org.apache.flink.api.common.typeutils.TypeSerializer;
+import org.apache.flink.api.java.functions.KeySelector;
+import org.apache.flink.streaming.api.datastream.DataStream;
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.table.catalog.ResolvedSchema;
+import org.apache.flink.table.connector.ChangelogMode;
+import org.apache.flink.table.runtime.typeutils.ExternalTypeInfo;
+import org.apache.flink.table.types.DataType;
+import org.apache.flink.types.Row;
+
+import com.alibaba.feathub.flink.udf.processfunction.SlidingWindowKeyedProcessFunction;
+
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+/** Utility methods to apply sliding windows. */
+public class SlidingWindowUtils {
+
+    /**
+     * Apply sliding window with the given step size to the given {@link Table} and {@link
+     * AggregationFieldsDescriptor}. The {@link AggregationFieldsDescriptor} describes how to
+     * compute each field. It includes the field name and data type of the input and output, the
+     * size of the sliding window under which the aggregation performs, and the aggregation
+     * function, e.g. SUM, AVG, etc.
+     *
+     * <p>The {@link SlidingWindowKeyedProcessFunction} is optimized to reduce the state usage when
+     * computing aggregations under different sliding window sizes.
+     *
+     * @param tEnv The StreamTableEnvironment of the table.
+     * @param table The input table.
+     * @param keyFieldNames The names of the group by keys for the sliding window.
+     * @param rowTimeFieldName The name of the row time field.
+     * @param stepSizeMs The step size of the sliding window in milliseconds.
+     * @param aggregationFieldsDescriptor The descriptor of the aggregation field in the sliding
+     *     window.
+     */
+    public static Table applySlidingWindowKeyedProcessFunction(
+            StreamTableEnvironment tEnv,
+            Table table,
+            String[] keyFieldNames,
+            String rowTimeFieldName,
+            long stepSizeMs,
+            AggregationFieldsDescriptor aggregationFieldsDescriptor) {
+        final ResolvedSchema resolvedSchema = table.getResolvedSchema();
+        DataStream<Row> rowDataStream =
+                tEnv.toChangelogStream(
+                        table,
+                        Schema.newBuilder().fromResolvedSchema(resolvedSchema).build(),
+                        ChangelogMode.all());
+
+        final TypeSerializer<Row> rowTypeSerializer =
+                ExternalTypeInfo.<Row>of(resolvedSchema.toPhysicalRowDataType())
+                        .createSerializer(null);
+        rowDataStream =
+                rowDataStream
+                        .keyBy(
+                                (KeySelector<Row, Row>)
+                                        value ->
+                                                Row.of(
+                                                        Arrays.stream(keyFieldNames)
+                                                                .map(value::getField)
+                                                                .toArray()))
+                        .process(
+                                new SlidingWindowKeyedProcessFunction(
+                                        aggregationFieldsDescriptor,
+                                        rowTypeSerializer,
+                                        rowTimeFieldName,
+                                        stepSizeMs))
+                        .returns(
+                                ExternalTypeInfo.of(
+                                        DataTypes.ROW(
+                                                getTableFields(
+                                                        keyFieldNames,
+                                                        rowTimeFieldName,
+                                                        aggregationFieldsDescriptor,
+                                                        resolvedSchema))));
+
+        table =
+                tEnv.fromDataStream(
+                        rowDataStream,
+                        getResultTableSchema(
+                                resolvedSchema,
+                                aggregationFieldsDescriptor,
+                                rowTimeFieldName,
+                                keyFieldNames));
+        for (AggregationFieldsDescriptor.AggregationFieldDescriptor aggregationFieldDescriptor :
+                aggregationFieldsDescriptor.getAggFieldDescriptors()) {
+            table =
+                    table.addOrReplaceColumns(
+                            $(aggregationFieldDescriptor.outFieldName)
+                                    .cast(aggregationFieldDescriptor.outDataType)
+                                    .as(aggregationFieldDescriptor.outFieldName));
+        }
+        return table;
+    }
+
+    private static List<DataTypes.Field> getTableFields(
+            String[] keyFieldNames,
+            String rowTimeFieldName,
+            AggregationFieldsDescriptor aggregationFieldsDescriptor,
+            ResolvedSchema resolvedSchema) {
+        List<DataTypes.Field> keyFields =
+                Arrays.stream(keyFieldNames)
+                        .map(
+                                fieldName ->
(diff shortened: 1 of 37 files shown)


Sxnan:
@lindong28 Could you take a look at this PR?

python/feathub/processors/flink/table_builder/aggregation_utils.py
```diff
@@ -64,6 +61,38 @@ def from_feature(feature: Feature) -> "AggregationFieldDescriptor":
         )
 
 
+class SlidingWindowAggregationFieldDescriptor(AggregationFieldDescriptor):
```
lindong28, line 64: The key semantics of this class is to represent `AggregationFieldDescriptor` with specific window size, and the same optimization can be applied to over window, right?

Can you rename the class to match its semantics?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/MultiSizeSlidingWindowUtils.java
```diff
@@ -0,0 +1,176 @@
...
+
+import java.util.Arrays;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static org.apache.flink.table.api.Expressions.$;
+
+/** Utility method to apply sliding windows with different size. */
+public class MultiSizeSlidingWindowUtils {
```
lindong28, line 40: How about naming this class as `SlidingWindowUtils` to match `sliding_window_utils.py`? The method name can reflect whether it is used for sliding window transforms of the same window size.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/aggregation/AggregationFunction.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.alibaba.feathub.flink.udf.aggregation;
+
+import org.apache.flink.table.types.DataType;
+
+import java.io.Serializable;
+
+/** Interface of aggregation function. */
+public interface AggregationFunction<IN_T, OUT_T> extends Serializable {
```
lindong28, line 24: Can we make the names/locations of the existing classes (e.g. TimeWindowedFirstValueAggFunc) consistent with classes newly added in this PR?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/aggregation/AggregationFunction.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package com.alibaba.feathub.flink.udf.aggregation;
+
+import org.apache.flink.table.types.DataType;
+
+import java.io.Serializable;
+
+/** Interface of aggregation function. */
```
lindong28, line 23: Can you add the doc to explain the difference between this class and `AbstractTimeWindowedAggFunc`?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/MultiSizeSlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,181 @@
...
+ * should be applied after the input has been aggregated with a tumble window with size equals to
+ * the step size of sliding window. With this process function, we only need to keep the rows for
+ * the maximum window size, instead of rows for each sliding window.
+ *
+ * <p>The ProcessFunction assumes that: 1. rows of each key are ordered by the row time 2. row time
+ * attribute of rows with the same key are all distinct. The assumptions hold true after applying
+ * the tumbling window aggregation whose window size is the same as the step size of the {@link
+ * MultiSizeSlidingWindowKeyedProcessFunction} to be applied.
+ */
+public class MultiSizeSlidingWindowKeyedProcessFunction
```
lindong28, line 47: Can we make the class package private (i.e. remove the `public` keyword) since it is supposed to be used only by MultiSizeSlidingWindowUtils?

Can we reorganize the files so that the public classes (e.g. MultiSizeSlidingWindowUtils, PostSlidingWindowUtils) are at higher level than package private classes?

lindong28 (commented):

Sxnan:

lindong28 (commented):

Can you also update the PR description to match its motivation.

Sxnan:

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,190 @@
...
+
+        out.collect(
+                Row.join(
+                        ctx.getCurrentKey(),
+                        aggResultRow,
+                        Row.of(Instant.ofEpochMilli(timestamp))));
+    }
+
+    /** The state of {@link SlidingWindowKeyedProcessFunction}. */
+    public static class MultiWindowSizeState {
```
lindong28, line 141: Can we make this class private?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,190 @@
...
+        private final MapState<Long, Row> rowState;
+        private final Map<Row, LinkedList<Long>> orderedTimestampMap;
+
+        private MultiWindowSizeState(MapState<Long, Row> mapState) {
+
+            this.rowState = mapState;
+            this.orderedTimestampMap = new HashMap<>();
+        }
+
+        public static MultiWindowSizeState buildMultiWindowSizeState(
```
lindong28, line 151: Would it be simpler to name it `create(...)`?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,190 @@
...
+
+        public void pruneRow(Row key, long lowerBound) throws Exception {
+            LinkedList<Long> orderedTimestamp = orderedTimestampMap.get(key);
+            if (orderedTimestamp == null) {
+                return;
+            }
+            final Iterator<Long> iterator = orderedTimestamp.iterator();
+            while (iterator.hasNext()) {
+                final long cur = iterator.next();
+                if (cur >= lowerBound) {
```
lindong28, line 182: Should we also remove the entry if `cur == lowerBound`?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public void processElement(
+            Row row, KeyedProcessFunction<Row, Row, Row>.Context ctx, Collector<Row> out)
+            throws Exception {
+        final long rowTime = ((Instant) row.getFieldAs(rowTimeFieldName)).toEpochMilli();
+        long triggerTime = rowTime;
+        while (triggerTime <= rowTime + aggregationFieldsDescriptor.getMaxWindowSizeMs()) {
+            ctx.timerService().registerEventTimeTimer(triggerTime);
```
lindong28, line 89: Since we already know that the inputs are ordered by time, would it be simpler and more performant to compute the results every time we receive an event without using timer?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -0,0 +1,190 @@
...
+            long timestamp,
+            KeyedProcessFunction<Row, Row, Row>.OnTimerContext ctx,
+            Collector<Row> out)
+            throws Exception {
+        state.pruneRow(
+                ctx.getCurrentKey(), timestamp - aggregationFieldsDescriptor.getMaxWindowSizeMs());
+        aggregationFieldsDescriptor.getAggFieldDescriptors().forEach(d -> d.aggFunc.reset());
+
+        boolean hasRow = false;
+        for (long rowTime : state.orderedTimestampMap.get(ctx.getCurrentKey())) {
```
lindong28, line 106: Suppose window_size=3600 sec and step_size=1 sec, then we need to 3600 aggregation-operation values at each step.

Is it possible to use retraction so that we only need 1 aggregation operation at each step?

Which approach does Flink window uses? If it is the latter, then we might introduce regression here.

lindong28 (commented):

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
```diff
@@ -0,0 +1,186 @@
...
+                    aggregationFieldDescriptor.outFieldName,
+                    aggregationFieldDescriptor.aggFunc.getResultDatatype());
+        }
+
+        builder.column(rowTimeFieldName, getDataType(resolvedSchema, rowTimeFieldName));
+
+        // Records are ordered by row time after sliding window.
+        builder.watermark(
+                rowTimeFieldName,
+                String.format("`%s` - INTERVAL '0.001' SECONDS", rowTimeFieldName));
```
lindong28, line 183: Since we have chosen to use `window_time` emitted by the tumble window function, and window_time = window_end - 1 ms, do we still need to subtract 1 ms here?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
Sxnan: We have to compute the result using timers because even after the tumbling window operation, it doesn't guarantee that there is a row at every step, and we need to perform aggregation at every step.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
Sxnan: Unfortunately, I think we still need to subtract 1 ms here. The main reason is not that the orders of the rows but that rows can have the same timestamp. [reference](https://stackoverflow.com/questions/54155126/does-apache-flink-support-multiple-events-with-the-same-timestamp)

> A Watermark is an assertion that all events that follow will have timestamps greater than the watermark, so this does mean that you cannot safely emit a Watermark t for a stream element at time t, unless the timestamps in the stream are strictly monotonically increasing -- which is not the case if there are multiple events with the same timestamp. This is why the AscendingTimestampExtractor produces watermarks equal to currentTimestamp - 1, and you should do the same.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
Sxnan: I think the native Flink window operator uses a slightly different approach with the same time complexity. For each record, it is assigned and aggregated for every window it belongs to. [reference](https://stackoverflow.com/questions/64566540/apache-flink-incremental-window-computation). There is an [opensource project](https://tu-berlin-dima.github.io/scotty-window-processor/) that aims to solve the problem. 

I think we can certainly do the optimization in Feathub sliding window so I added a TODO.

Sxnan:
@lindong28 Thank you for your review! I updated the PR. Could you please take another look?

lindong28:
Thanks for the update! LGTM.

PR #51, opened 2022-12-13
Raises exception when sinking csv files to s3 using Flink processor

jiangxin369:

Since Flink has a bug when sinking CSV files to s3, see [FLINK-28513](https://issues.apache.org/jira/browse/FLINK-28513). Feathub should check for this during compile time to avoid unclear exceptions.

## Brief change log

  - Raises exception when sinking csv files to s3 using Flink processor

  - The public API: (no)

## Documentation


lindong28:
Thanks for the PR. LGTM.

PR #53, opened 2022-12-17
Refactor OnlineStoreSource into MemoryStoreSource and OnlineStoreSink into MemoryStoreSink

lindong28:

The purpose of this PR is:
- Refactor OnlineStoreSource into MemoryStoreSource and OnlineStoreSink into MemoryStoreSink.
- Remove `stores` from the constructor of FeatureService and Processor classes.
- Improve the instructions in README to be more readable.
- Improve the nightly build script to be more readable.

  - The public API: yes

## Documentation


lindong28:

README.md
```diff
@@ -179,35 +179,46 @@ f_lower_case_name = Feature(
...
-for installation guidelines. 
+```bash
+$ docker image pull redis:latest
+$ docker image pull confluentinc/cp-kafka:5.4.3
+```
+
+3. Increase open file limit.
+
+```bash
+$ ulimit -n 1024
```
yunfengzhou-hub, line 198: Since in #45 we have greatly reduced the Flink environments created during tests, increasing the open file limit might be unnecessary.

.github/workflows/build-and-test.yaml
```diff
@@ -41,20 +41,23 @@ jobs:
...
           protoc -I=python/feathub/common/protobuf/ --python_out=python/feathub/common/protobuf/ value.proto
           
-          python -m pip install ./python
           python -m pip install -r python/dev-requirements.txt
+          python -m pip install ./python
           
+          # Re-install black here because its pip dependency conflicts with
+          # apache-flink 1.15.2.
+          python -m pip install black==22.3.0
           python -m black --check python
```
yunfengzhou-hub, line 56: nit: use `./python` instead of `python` here.

README.md
```diff
@@ -222,13 +233,13 @@ before uploading PRs for review.
 
 ```bash
 # Format python code
-$ python -m black python
+$ python -m black ./python
```
yunfengzhou-hub, line 236: Since we re-installed black dependency in `build-and-test.yaml`, we may need to re-install it here as well.

python/feathub/online_stores/memory_online_store.py
```diff
@@ -138,3 +175,13 @@ def get(
...
+
+    def reset(self) -> None:
+        self.table_infos = {}
+
+    @staticmethod
+    def get_instance() -> MemoryOnlineStore:
+        if MemoryOnlineStore.INSTANCE is None:
+            MemoryOnlineStore.INSTANCE = MemoryOnlineStore()
+
+        return MemoryOnlineStore.INSTANCE
```
yunfengzhou-hub, line 187: In the current implementation, Feathub would create a `MemoryOnlineStore` every time this method is invoked, while this PR proposes to create a singleton instance. Would this affect the correctness of feature values when there are multiple feature services or processors accessing the same table name in the memory store?

python/feathub/online_stores/memory_online_store.py
```diff
@@ -138,3 +175,13 @@ def get(
         if table_info.timestamp_field is not None and not include_timestamp_field:
             features = features.drop(columns=[field_to_drop])
         return features
+
+    def reset(self) -> None:
```
yunfengzhou-hub, line 179: Do you think it would be better if we add `open()` & `close()`, or `initialize()` and `cleanup()` method to `FeathubClient` or `FeatureService` class? These methods would help users setup resources beforehand and closes them when they are sure they do not need them anymore. And we can invoke `MemoryOnlineStore.reset()` in those methods.

If not, this method is currently only used in tests. It might be better to create a utility class for this function in test_utils, instead of making it a public API.

python/feathub/feature_tables/sinks/memory_store_sink.py
```diff
@@ -16,27 +16,24 @@
 from feathub.feature_tables.sinks.sink import Sink
 
 
-class OnlineStoreSink(Sink):
+class MemoryStoreSink(Sink):
```
yunfengzhou-hub, line 19: Shall we rename this class to `MemorySink` instead of `MemoryStoreSink`, in correspondence to `RedisSink` and `KafkaSink`?

yunfengzhou-hub (commented):
Thanks for the PR! Left some comments as below.

README.md
lindong28: Due to bug in pyflink, the number of open file descriptor will increase as the number of test classes increase. I tested locally and found that FeatHub python tests will fail if number of open file limit is 256. Setting this limit to 1024 would allow tests to pass.

.github/workflows/build-and-test.yaml
lindong28: It is fixed now.

README.md
lindong28: Discussed offline. We agree it is simpler to keep it as is and users should know how to install black if it is not find in the python library.

python/feathub/online_stores/memory_online_store.py
lindong28: MemoryOnlineStore is only needed in tests. We agree there is no correctness issue in the current design.

python/feathub/online_stores/memory_online_store.py
lindong28: MemoryOnlineStore is only needed in tests. We agree it is simpler not to add those methods in FeathubClient or FeatureService.

RedisClient can be closed when FeatHubClient is destructed.

python/feathub/feature_tables/sinks/memory_store_sink.py
lindong28: It is not clear whether MemorySink refers to a queue or kv store. So it seems more readable to name it `MemoryStoreSink`.

Redis is typically used as kv store. And other projects such as Apache Bahir name it RedisSink. So it seems better to keep the same name here.

lindong28:
@yunfengzhou-hub Thanks for the review. The PR has been updated and all comments are addressed.

yunfengzhou-hub (commented):
Thanks for the update! It LGTM.

PR #56, opened 2022-12-29
Update LocalProcessor OverWindow and AggFunc

Sxnan:

This PR updates LocalProcessor OverWindow to support over-window transformation with filter and limit. It also adds the implementation of all the missing AggFunc.

## Brief change log

- Update LocalProcessor OverWindow to support over-window transformation with filter and limit.
- Add implementation of AggFunc (FIRST_VALUE, LAST_VALUE, ROW_NUMBER, VALUE_COUNTS)

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
+++ b/python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
@@ -70,7 +70,7 @@ class LocalAstEvaluator(AbstractAstEvaluator):
             return left_value > right_value
         elif ast.op_type == ">=":
             return left_value >= right_value
-        elif ast.op_type == "==":
+        elif ast.op_type == "=":
             return left_value == right_value
         elif ast.op_type == "<>":
             return left_value != right_value
--- a/docs/feathub_expression.md
+++ b/docs/feathub_expression.md
@@ -40,7 +40,7 @@ Comparison functions take numeric values as inputs and outputs a boolean value.
 | x >= y | Returns TRUE iff `x >= y`. Returns NULL if in case of error. |
 | x < y | Returns TRUE iff `x < y`. Returns NULL if in case of error. |
 | x <= y | Returns TRUE iff `x <= y`. Returns NULL if in case of error. |
-| x == y | Returns TRUE iff `x` is equal to `y`. Returns NULL if in case of error. |
+| x = y | Returns TRUE iff `x` is equal to `y`. Returns NULL if in case of error. |
 | x <> y | Returns TRUE iff `x` is not equal to `y`. Returns NULL if in case of error. |
 
 ## Logical Functions
--- a/docs/how_to_guides/alibaba_cloud_deployment.md
+++ b/docs/how_to_guides/alibaba_cloud_deployment.md
@@ -19,7 +19,7 @@ the script.
 $ bash tools/vvp-dep/build-vvp-dep.sh
 ```
 
-The dependencies zip will be available at `tools/vvp-dep/deps.zip`. 
+The dependencies zip will be available at `tools/cli-deps/deps.zip`. 
 
 You can modify the script to include any additional Python dependencies to the 
 `deps.zip`. Or you can package your own zip that only includes the additional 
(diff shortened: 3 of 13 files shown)


Sxnan:

python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
```diff
@@ -70,7 +70,7 @@ def eval_compare_op(self, ast: CompareOp, variables: Optional[Dict]) -> Any:
             return left_value > right_value
         elif ast.op_type == ">=":
             return left_value >= right_value
-        elif ast.op_type == "==":
+        elif ast.op_type == "==" or ast.op_type == "=":
```
lindong28, line 73: Can you explain why we need to support both `==` and `=` for the same logic expression? Is there any popular DSL language `e.g. SQL` that does this?

python/feathub/processors/local/tests/test_local_processor_over_window.py
```diff
@@ -0,0 +1,832 @@
...
+from feathub.common.test_utils import LocalProcessorTestCase
+from feathub.common.types import Int64, String, Float64, MapType
+from feathub.feature_views.derived_feature_view import DerivedFeatureView
+from feathub.feature_views.feature import Feature
+from feathub.feature_views.transforms.over_window_transform import OverWindowTransform
+from feathub.feature_views.transforms.python_udf_transform import PythonUdfTransform
+from feathub.online_stores.memory_online_store import MemoryOnlineStore
+
+
+class FlinkTableBuilderOverWindowTransformTest(LocalProcessorTestCase):
```
lindong28, line 28: Update the class name as appropriate?

Can you also update the file name of this and existing files so that it is clear what is the relationship/difference between this test and `test_local_processor.py`?

python/feathub/processors/local/tests/test_local_processor_over_window.py
```diff
@@ -0,0 +1,832 @@
...
+        with self.assertRaises(ValueError):
+            Feature(
+                name="feature_1",
+                dtype=Int64,
+                transform=OverWindowTransform(
+                    "cost", "unsupported_agg", window_size=timedelta(days=2)
+                ),
+            )
+
+    def test_out_of_order(self):
```
lindong28, line 57: The name of this test seems inconsistent with other tests in this class.

python/feathub/processors/local/tests/test_local_processor_over_window.py
```diff
@@ -0,0 +1,832 @@
...
+from feathub.feature_views.feature import Feature
+from feathub.feature_views.transforms.over_window_transform import OverWindowTransform
+from feathub.feature_views.transforms.python_udf_transform import PythonUdfTransform
+from feathub.online_stores.memory_online_store import MemoryOnlineStore
+
+
+class FlinkTableBuilderOverWindowTransformTest(LocalProcessorTestCase):
+    def setUp(self):
+        super().setUp()
+        self.input_data = pd.DataFrame(
```
lindong28, line 31: By design of FeatHub, users should be able to switch between LocalProcessor and FlinkProcessor by only changing the configuration. Thus this test should be able to share a lot of code with `FlinkTableBuilderOverWindowTransformTest`.

Could you refactor the code to take advantage of design so that we re-use a lot of test code between LocalProcessor, FlinkProcessor and SparkProcessor? This can also help us catch and fix cases that does not follow this design pattern.

lindong28 (commented):

python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
Sxnan: I think SQL, both [FlinkSQL](https://nightlies.apache.org/flink/flink-docs-release-1.15/docs/dev/table/functions/systemfunctions/#comparison-functions) or [MySQL](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html) use and only use `=` for equal comparison. `=` is added so that it is compatible with SQL.

For `==`, I think we can keep it as it is not used in SQL. What do you think?

Sxnan:
@lindong28 Thanks for your review! I updated the PR and all the comments are addressed. Could you please take another look?

python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
lindong28: Do you mean we should keep `==`?

Since we want to be compatible with Flink SQL, would it be simpler to remove `=` from the FeatHub expression language so that we have a consistent way to express the equal logic?

Sxnan: I agree it is cleaner to only support `=` and remove `==` in Feathub DSL.

Sxnan:

lindong28:
Thanks for the update! LGTM.

PR #59, opened 2023-01-12
Unify tests for different Processors

yunfengzhou-hub:

This pull request restructures existing integration tests for Feathub Processors so that different processors or deployment modes can be verified against the same test cases.

## Brief change log

This pull request mainly adds the following test classes:

- Adds `FeathubITTestBase`. 
  - It is the base class for all Feathub Processor's integration tests. 
  - It provides several functions or tools that are supposed to be useful in most test cases.
- Adds the following abstract classes that inherit `FeathubITTestBase` and hold test cases for corresponding functionalities.
  - Classes that verify FeatureViews.
    - `ExpressionTransformITTest`
    - `OverWindowTransformITTest`
    - `SlidingWindowTransformITTest`
    - `JoinTransformITTest`
    - `PythonUDFTransformITTest`
  - Classes that verify sources and sinks.
    - `DataGenSourceITTest`
    - `FileSystemSourceSinkITTest`
    - `KafkaSourceSinkITTest`
    - `PrintSinkITTest`
    - `RedisSourceSinkITTest`
  - Classes that verify other Feathub public APIs.
    - `GetFeaturesITTest`
    - `OnlineFeaturesITTest`
- Adds the following abstract classes that inherit `ProcessorTestBase`, provide corresponding FeathubClient instances, and execute the test cases defined in the classes above with the FeathubClient.
  - `LocalProcessorITTest`
  - `FlinkProcessorITTest`
  - `SparkProcessorITTest`

This PR also completes the following bugfixes.

- Supports `FlinkTable#to_pandas` when FlinkProcessor is created in CLI mode.
- Supports removing source field in `SparkProcessor#get_table` when the `FeatureTable` contains `DerivedFeatureView`.
- Supports overwriting existing files with `FileSystemSink` in `SparkProcessor`.
- Fixes the description of the parameter `allow_overwrite`.

This pull request has the following limitations that are planned to be covered by follow-up pull requests. These limitations have been recorded as TODOs in the codes if needed.

- This pull request only migrates existing test cases and deduplicates them. It does not check whether the existing test cases have been reasonable enough.
- Apart from the bugfixes mentioned above, this pull request does not fix other bugs discovered in this restructuring process. If some test case fails due to a potential bug in a Processor subclass, this pull request ignores the test case for now.
- This pull request only restructures tests in `feathub.processors` module, leaving alone other test classes that may have the same need, like `feathub/online_stores/tests/test_redis_client.py`.

  - The public API: no

## Documentation

  - Does this pull request introduce a new feature? No


python/feathub/processors/tests/feathub_test_base.py
```diff
@@ -0,0 +1,157 @@
...
+                ["Alex", 600, 800, "2022-01-03 08:06:00"],
+            ],
+            columns=["name", "cost", "distance", "time"],
+        )
+
+        schema = self._create_input_schema()
+
+        return input_data, schema
+
+    def _create_input_data_and_schema_with_millis_time_span(self):
```
lindong28, line 126: It seems inconsistent that `_create_input_schema` is marked as static but this method is not.

python/feathub/processors/tests/feathub_test_base.py
```diff
@@ -0,0 +1,157 @@
...
+
+    @abstractmethod
+    def get_client(self) -> FeathubClient:
+        """
+        Returns a FeathubClient instance for test cases.
+        """
+        pass
+
+    @staticmethod
+    def _get_local_client(processor_config: Dict) -> FeathubClient:
```
lindong28, line 61: Since this method can be called by subclasses, it seems better to rename this method as `get_local_client()`.

Same for other methods that can be called outside this class.

python/feathub/processors/tests/mixed_transform_test_base.py
```diff
@@ -0,0 +1,781 @@
...
+from feathub.processors.tests.feathub_test_base import FeathubTestBase
+from feathub.processors.tests.sliding_window_transform_test_base import (
+    ENABLE_EMPTY_WINDOW_OUTPUT_SKIP_SAME_WINDOW_OUTPUT,
+    DISABLE_EMPTY_WINDOW_OUTPUT_WITHOUT_SKIP_SAME_WINDOW_OUTPUT,
+    ENABLE_EMPTY_WINDOW_OUTPUT_WITHOUT_SKIP_SAME_WINDOW_OUTPUT,
+)
+from feathub.table.schema import Schema
+
+
+class MixedTransformTestBase(FeathubTestBase):
```
lindong28, line 40: It is probably better to move tests in this class to class with more informative names (e.g. OverWindowTransformTestBase). It is OK for a test in `OverWindowTransformTestBase` to also use other Transforms. And some information is better than no information.

python/feathub/processors/flink/tests/test_flink_processor.py
```diff
@@ -11,311 +11,263 @@
...
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        if "PYFLINK_GATEWAY_DISABLED" in os.environ:
+            os.environ.pop("PYFLINK_GATEWAY_DISABLED")
+
+    def get_client(self) -> FeathubClient:
+        return self.client
+
+    def test_read_write(self) -> None:
```
lindong28, line 107: Remove the `None` for consistency?

python/feathub/processors/tests/datagen_source_test_base.py
```diff
@@ -11,24 +11,28 @@
...
 
 
-class DataGenUtilsTest(FlinkTableBuilderTestBase):
+class DataGenSourceTestBase(FeathubTestBase):
+    """
+    Base class that provides test cases to verify DataGenSource.
+    """
+
+    @abstractmethod
+    def get_client(self) -> FeathubClient:
```
lindong28, line 33: Would it be simpler to remove this method here?

Same for other test bases.

lindong28 (commented):

lindong28:
Thanks for the update! LGTM.

PR #61, opened 2023-01-18
Fix FlinkProcessor IT failures by reorg file structure

yunfengzhou-hub:

This pull request mainly provides a temporary fix for the failures in FlinkProcessor's integration tests.

## Brief change log

- Moves `test_flink_processor.py` from `python/feathub/processors/flink/tests` to `python/feathub/processors/flink/table_builder/tests`
- Moves `test_redis_client.py` from `python/feathub/online_stores/tests` to `python/feathub/processors/flink/tests`
- Adds cleanup of java gateway in `test_redis_client.py`.
- Removes `GetFeaturesITTest` from `FlinkProcessorITTest`
  - flink processor has not been tested against these test cases before commit 34fad8.
- Moves container setup/teardown process to `setUpClass()`/`tearDownClass()` methods in kafka and redis it test class.
- Makes `xxxProcessorITTest` executes base class's `setUp()`/`tearDown()`/`setUpClass()`/`tearDownClass()` methods.

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/feature_tables/tests/test_kafka_source_sink.py
+++ b/python/feathub/feature_tables/tests/test_kafka_source_sink.py
@@ -54,6 +54,17 @@ class KafkaSourceTest(unittest.TestCase):
 class KafkaSourceSinkITTest(ABC, FeathubITTestBase):
     kafka_container = None
 
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        cls.kafka_container = KafkaContainer()
+        cls.kafka_container.start()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        super().tearDownClass()
+        cls.kafka_container.stop()
+
     def test_kafka_source_sink(self):
         input_data = pd.DataFrame(
             [
@@ -77,7 +88,7 @@ class KafkaSourceSinkITTest(ABC, FeathubITTestBase):
         # Consume data with kafka source
         source = KafkaSource(
             "kafka_source",
-            bootstrap_server=self._get_kafka_bootstrap_servers(),
+            bootstrap_server=self.kafka_container.get_bootstrap_server(),
             topic=topic_name,
             key_format="json",
             value_format="json",
@@ -125,7 +136,7 @@ class KafkaSourceSinkITTest(ABC, FeathubITTestBase):
         # Consume data with kafka source
         source = KafkaSource(
             "kafka_source",
-            bootstrap_server=self._get_kafka_bootstrap_servers(),
+            bootstrap_server=self.kafka_container.get_bootstrap_server(),
             topic=topic_name,
             key_format="json",
             value_format="json",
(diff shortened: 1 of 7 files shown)


python/feathub/processors/flink/tests/test_redis_client.py
```diff
@@ -50,6 +54,13 @@ def setUpClass(cls) -> None:
     def tearDownClass(cls) -> None:
         super().tearDownClass()
         cls.redis_container.stop()
+        # clean up the java_gateway so that it won't affect other tests.
+        with java_gateway._lock:
+            if java_gateway._gateway is not None:
```
lindong28, line 59: It seems inconsistent to explicitly tear down java_gateway here while we don't have code that explicitly setup  java_gateway in `setUpClass`.

Is this the expected way to use `pyflink.java_gateway` or is this due to bug? If it is due to bug, can you provide the JIRA ticket here and add TODO to remove this code snippet?

python/feathub/processors/flink/table_builder/tests/test_flink_processor.py
```diff
@@ -400,8 +402,6 @@ def setUp(self):
     def tearDown(self) -> None:
         MemoryOnlineStore.get_instance().reset()
         shutil.rmtree(self.temp_dir, ignore_errors=True)
-        KafkaSourceSinkITTest.tearDown(self)
```
lindong28, line 403: Would it be simpler to have a for-loop that enumerates all superclass and call `teardown()` for all superclasses who have `teardown()`?

If we agree to do this, we can either address it in this PR or add a TODO and work on this in a followup PR (without removing the code here).

Same for `setUp()` and for other processor IT tests.

python/feathub/feature_tables/tests/test_kafka_source_sink.py
```diff
@@ -102,6 +102,8 @@ def test_kafka_source_sink(self):
         )
         self.assertTrue(expected_result_df.equals(result_df))
 
+        self._tear_down_kafka()
```
lindong28, line 105: Would it be simpler to teardown `cls.kafka_container` in `KafkaSourceSinkITTest#tearDownClass`?

lindong28 (commented):

python/feathub/tests/feathub_it_test_base.py
```diff
@@ -106,6 +106,38 @@ def create_file_source(
...
+    def invoke_base_class_setup(self):
+        self._get_base_test_class().setUp(self)
+
+    def invoke_base_class_teardown(self):
+        self._get_base_test_class().tearDown(self)
+
+    def _get_base_test_class(self) -> Type[unittest.TestCase]:
+        for base_class in self.__class__.__bases__:
+            method_list = [
+                func for func in dir(base_class) if callable(getattr(base_class, func))
```
lindong28, line 132: Instead of enumerating all methods in all base classes, is it possible to use `unittest.TestCase.id()` to get the base class name?

If it does not work, can we setup a hash table to map test name to base class name, instead of enumerating all methods for each test?

yunfengzhou-hub: `unittest.TestCase.id()` would return the current test class instead of the base class containing the current test method, thus we cannot directly use it in this situation. I'll add a hash table to hold the mapping of test names to base class names.

python/feathub/tests/feathub_it_test_base.py
```diff
@@ -106,6 +111,47 @@ def create_file_source(
...
+        self._get_base_test_class().setUp(self)
+
+    def invoke_base_class_teardown(self):
+        self._get_base_test_class().tearDown(self)
+
+    def _get_base_test_class(self) -> Type[unittest.TestCase]:
+        if self._base_class_mapping is None:
+            self._base_class_mapping = dict()
+            for base_class in self.__class__.__bases__:
+                if not issubclass(base_class, unittest.TestCase):
```
lindong28, line 138: nits: it might be more readable to write it in this way

```
for func in dir(base_class):
    if not callable(getattr(base_class, func)) and func.startswith(
        TestLoader.testMethodPrefix
    ):
        continue
    if func in self._base_class_mapping:
        raise FeathubException(
            f"Duplicated test case name {func} detected in "
            f"integration test base class "
            f"{self._base_class_mapping[func]} and {base_class}."
        )
```

lindong28:
Thanks for the update. LGTM.

PR #67, opened 2023-02-03
Support LOWER, NULL, CASE syntax and filtering Derived/SlidingFeatureView in FlinkProcessor

yunfengzhou-hub:

This pull request adds support for several expression syntaxes and filtering in Derived/SlidingFeatureView in FlinkProcessor.

## Brief change log

- Support the following syntaxes in Feathub's parser and FlinkProcessor.
  - `LOWER(str)`
  - `CASE WHEN expr1 THEN expr2 [WHEN expr3 THEN expr4]* [ELSE expr5] END`
  - `x IS[ NOT] NULL`
- Add API to filter features in `DerivedFeatureView` and `SlidingFeatureView`, and add corresponding support in FlinkProcessor.
- Fix the keyword of logical operations in Feathub(`&&/||` -> `AND/OR`).

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? docs and python docs.

### DIFF
--- a/docs/feathub_expression.md
+++ b/docs/feathub_expression.md
@@ -34,14 +34,16 @@ Arithmetic functions take numeric values as inputs and outputs a numeric value.
 
 Comparison functions take numeric values as inputs and outputs a boolean value.
 
-| Function | Description |
-| --- | --- |
-| x > y | Returns TRUE iff `x > y`. Returns NULL if in case of error. |
-| x >= y | Returns TRUE iff `x >= y`. Returns NULL if in case of error. |
-| x < y | Returns TRUE iff `x < y`. Returns NULL if in case of error. |
-| x <= y | Returns TRUE iff `x <= y`. Returns NULL if in case of error. |
-| x = y | Returns TRUE iff `x` is equal to `y`. Returns NULL if in case of error. |
-| x <> y | Returns TRUE iff `x` is not equal to `y`. Returns NULL if in case of error. |
+| Function      | Description                                                              |
+|---------------|--------------------------------------------------------------------------|
+| x > y         | Returns TRUE iff `x > y`. Returns NULL in case of error.                 |
+| x >= y        | Returns TRUE iff `x >= y`. Returns NULL in case of error.                |
+| x < y         | Returns TRUE iff `x < y`. Returns NULL in case of error.                 |
+| x <= y        | Returns TRUE iff `x <= y`. Returns NULL in case of error.                |
+| x = y         | Returns TRUE iff `x` is equal to `y`. Returns NULL in case of error.     |
+| x <> y        | Returns TRUE iff `x` is not equal to `y`. Returns NULL in case of error. |
+| x IS NULL     | Returns TRUE iff `x` is NULL.                                            |
+| x IS NOT NULL | Returns TRUE iff `x` is not NULL.                                        |
 
 ## Logical Functions
 
@@ -55,7 +57,7 @@ Comparison functions take numeric values as inputs and outputs a boolean value.
 | Function             | Description                                                                               |
 |----------------------|-------------------------------------------------------------------------------------------|
 | CAST(x AS DTYPE)     | Returns a new value being casted to a the given DTYPE. Throws exception in case of error. |
-| TRY_CAST(x AS DTYPE) | Returns a new value being casted to a the given DTYPE. Returns NULL if in case of error.  |
+| TRY_CAST(x AS DTYPE) | Returns a new value being casted to a the given DTYPE. Returns NULL in case of error.     |
 
 ### Data Types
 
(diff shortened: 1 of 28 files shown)


python/feathub/feature_views/tests/test_sliding_feature_view.py
```diff
@@ -404,3 +410,49 @@ def test_build_with_config(self):
...
+            dtype=Int64,
+            transform=SlidingWindowTransform(
+                expr="cost",
+                agg_func="SUM",
+                window_size=timedelta(days=2),
+                step_size=timedelta(days=1),
+            ),
+        )
+
+        expected_result_df = pd.DataFrame(
```
lindong28, line 431: Would it be better to put `expected_result_df` after `result_df = ...` than inserting it in the middle of the code that defines `features`?

python/feathub/feature_views/transforms/tests/test_expression_transform.py
```diff
@@ -172,3 +172,81 @@ def test_unix_timestamp_with_timezone(self):
...
+            lambda row: str(row["name"]).lower(),
+            axis=1,
+        )
+        expected_result_df = expected_result_df.sort_values(by=["time"]).reset_index(
+            drop=True
+        )
+
+        self.assertTrue(expected_result_df.equals(result_df))
+
+    def test_case(self):
```
lindong28, line 212: Can we also add a test that covers the use of `else`?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -52,13 +53,19 @@ def __init__(
                          this list and the features in the source table.
         :param keep_source_fields: True iff all fields in the source table should be
                                    included in this table.
+        :param filter_expr: Optional. If it is not None, it represents a Feathub
```
lindong28, line 56: Instead of saying `match the filter expression`, it might be more informative to say something like `the filter expression evaluates to True`.

How about this statement:

Optional. If it is not None, it represents a Feathub expression which evaluates to a boolean value. The filter expression is evaluated after other transformations in the feature view, and only those rows which evaluate to True will be outputted by the feature view.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -164,6 +170,8 @@ def __init__(
                 "is forbidden."
             )
 
+        self.filter_expr = filter_expr
```
lindong28, line 173: Would it be more readable to put it right after `super()` before we apply function like `_validate()`? Also note that functions such as `_validate()` in general might need to check arbitrary properties of this instance.

python/feathub/processors/flink/ast_evaluator/functions.py
```diff
@@ -16,7 +16,18 @@
...
+is the same between Feathub and in Flink.
+"""
+_functions_with_equal_name = {"LOWER"}
+
+
 def evaluate_function(func_name: str, args: List[Any]) -> str:
-    if func_name.upper() == "UNIX_TIMESTAMP" and len(args) > 1:
-        args[1] = to_java_date_format(args[1])
-    return f"{func_name}({', '.join(args)})"
+    if func_name.upper() in _functions_with_equal_name:
```
lindong28, line 27: It might be simpler to consistently make func_name upper-case in `feathub.dsl.ast.FuncCallOp` than having each processor make this change when reading the function name.

lindong28 (commented):

docs/feathub_expression.md
```diff
@@ -34,28 +34,30 @@ Arithmetic functions take numeric values as inputs and outputs a numeric value.
...
 
 ## Logical Functions
 
-| Function        | Description                                     |
-|-----------------|-------------------------------------------------|
-| bool1 OR bool2  | Returns TRUE if bool1 is TRUE or bool2 is TRUE. |
-| bool1 AND bool2 | Returns TRUE if bool1 and bool2 are both TRUE.  |
+| Function                 | Description                                     |
+|--------------------------|-------------------------------------------------|
+| bool1 &#124;&#124; bool2 | Returns TRUE if bool1 is TRUE or bool2 is TRUE. |
```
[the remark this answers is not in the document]
yunfengzhou-hub: `&#124;` stands for `|`. Given that "|" is a syntax in markdown table, we cannot directly write this character in the document[1]. This problem is not applicable now as this PR changes to `AND/OR` instead of `&&/||`.

[1] https://stackoverflow.com/questions/17319940/how-to-escape-a-pipe-char-in-a-code-statement-in-a-markdown-table

python/feathub/dsl/ast.py
```diff
@@ -166,3 +168,84 @@ def __init__(self, child: ExprAST) -> None:
...
+
+    def to_json(self) -> Dict:
+        return {"node_type": "NullNode"}
+
+
+class IsOp(ExprAST):
+    def __init__(
+        self, left_child: ExprAST, right_child: ExprAST, is_not: bool = False
+    ) -> None:
+        super().__init__(node_type="IsNullOp")
```
lindong28, line 185: Should we make the nodetype consistent with the class name?

python/feathub/dsl/ast.py
```diff
@@ -166,3 +168,84 @@ def __init__(self, child: ExprAST) -> None:
...
+        if not isinstance(right_child, NullNode):
+            raise FeathubException("IS/IS NOT can only be concatenated with NULL.")
+
+        self.left_child = left_child
+        self.right_child = right_child
```
lindong28, line 190: It seems simpler to remove `right_child` here.

python/feathub/feature_views/tests/test_derived_feature_view.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def test_filter_is_null(self):
+        source = self.create_file_source(self.input_data.copy())
+
+        def alex_to_null(row: pd.Series) -> Optional[str]:
+            if row["name"] == "Alex":
+                return None
+            return row["name"]
+
+        expected_result_df = pd.DataFrame(
```
lindong28, line 66: Would it be better to move this line to be close to the line 97?

lindong28 (commented):

PR #73, opened 2023-02-10
Make Feature dtype optional

### LINKED ISSUE #68: Make Feature#detype optional
FeatHub should be able to derive the feature's dtype automatically in common cases.

Sxnan:

This PR makes dtype of Feature optional. Fix #68.

## Brief change log

- Make dtype of Feature optional.

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/dsl/ast.py
+++ b/python/feathub/dsl/ast.py
@@ -13,9 +13,44 @@
 #  limitations under the License.
 import json
 from abc import ABC, abstractmethod
-from typing import List, Dict, Any
+from typing import List, Dict, Any, Optional, Callable
 
-from feathub.common.exceptions import FeathubException
+from feathub.common.exceptions import FeathubException, FeathubExpressionException
+from feathub.common.types import (
+    DType,
+    Int32,
+    Int64,
+    Float32,
+    Float64,
+    Bool,
+    String,
+    get_type_by_name,
+    from_python_type,
+)
+
+TYPE_PRECISION_RANK: List[DType] = [Float64, Float32, Int64, Int32]
+
+# A map from the Feathub built-in function name to a callable that is used to
+# evaluate the result data type of the function. The callable takes a list of
+# data types of arguments and returns the data type of the result.
+# TODO: Add a dedicated class to keep track of all the Feathub function.
+FUNCTION_RESULT_TYPE_EVALUATOR: Dict[str, Callable[[List[DType]], DType]] = {
+    "LOWER": lambda _: String,
+    "UNIX_TIMESTAMP": lambda _: Int64,
+}
+
+
+def _get_higher_precision_type(*dtype: DType) -> Optional[DType]:
+    res_type = dtype[0]
+
+    for t in dtype[1:]:
+        if res_type not in TYPE_PRECISION_RANK or t not in TYPE_PRECISION_RANK:
+            raise FeathubExpressionException(f"Illegal mixing of types: {dtype}")
+
+        if TYPE_PRECISION_RANK.index(res_type) > TYPE_PRECISION_RANK.index(t):
+            res_type = t
+
+    return res_type
 
 
 class ExprAST(ABC):
@@ -29,16 +64,39 @@ class ExprAST(ABC):
         """
         pass
 
+    @abstractmethod
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        pass
+
     def __str__(self) -> str:
         return json.dumps(self.to_json(), indent=2, sort_keys=True)
 
 
-class BinaryOp(ExprAST):
-    def __init__(self, op_type: str, left_child: ExprAST, right_child: ExprAST) -> None:
-        super().__init__(node_type="BinaryOp")
-        self.op_type = op_type
+class AbstractUnaryOp(ExprAST, ABC):
+    def __init__(self, node_type: str, child: ExprAST):
+        super().__init__(node_type)
+        self.child = child
+
+
+class AbstractBinaryOp(ExprAST, ABC):
+    def __init__(self, node_type: str, left_child: ExprAST, right_child: ExprAST):
+        super().__init__(node_type)
         self.left_child = left_child
         self.right_child = right_child
+
+
+class BinaryOp(AbstractBinaryOp):
+    def __init__(self, op_type: str, left_child: ExprAST, right_child: ExprAST) -> None:
+        super().__init__(
+            node_type="BinaryOp", left_child=left_child, right_child=right_child
+        )
+        self.op_type = op_type
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        left_type = self.left_child.eval_dtype(variable_types)
+        right_type = self.right_child.eval_dtype(variable_types)
+
+        return _get_higher_precision_type(left_type, right_type)
 
     def to_json(self) -> Dict:
         return {
@@ -49,12 +107,15 @@ class BinaryOp(ExprAST):
         }
 
 
-class CompareOp(ExprAST):
+class CompareOp(AbstractBinaryOp):
     def __init__(self, op_type: str, left_child: ExprAST, right_child: ExprAST) -> None:
-        super().__init__(node_type="CompareOp")
+        super().__init__(
+            node_type="CompareOp", left_child=left_child, right_child=right_child
+        )
         self.op_type = op_type
-        self.left_child = left_child
-        self.right_child = right_child
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        return Bool
 
     def to_json(self) -> Dict:
         return {
@@ -65,10 +126,12 @@ class CompareOp(ExprAST):
         }
 
 
-class UminusOp(ExprAST):
+class UminusOp(AbstractUnaryOp):
     def __init__(self, child: ExprAST) -> None:
-        super().__init__(node_type="UminusOp")
-        self.child = child
+        super().__init__(node_type="UminusOp", child=child)
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        return self.child.eval_dtype(variable_types)
 
     def to_json(self) -> Dict:
         return {
@@ -77,12 +140,15 @@ class UminusOp(ExprAST):
         }
 
 
-class LogicalOp(ExprAST):
+class LogicalOp(AbstractBinaryOp):
     def __init__(self, op_type: str, left_child: ExprAST, right_child: ExprAST) -> None:
-        super().__init__(node_type="LogicalOp")
+        super().__init__(
+            node_type="LogicalOp", left_child=left_child, right_child=right_child
+        )
         self.op_type = op_type.upper()
-        self.left_child = left_child
-        self.right_child = right_child
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        return Bool
 
     def to_json(self) -> Dict:
         return {
@@ -93,14 +159,16 @@ class LogicalOp(ExprAST):
         }
 
 
-class CastOp(ExprAST):
+class CastOp(AbstractUnaryOp):
     def __init__(
         self, child: ExprAST, type_name: str, exception_on_failure: bool = True
     ):
-        super().__init__(node_type="CastOp")
-        self.child = child
+        super().__init__(node_type="CastOp", child=child)
         self.type_name = type_name
         self.exception_on_failure = exception_on_failure
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        return get_type_by_name(self.type_name)
 
     def to_json(self) -> Dict:
         return {
@@ -116,6 +184,9 @@ class ValueNode(ExprAST):
         super().__init__(node_type="ValueNode")
         self.value = value
 
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        return from_python_type(type(self.value))
+
     def to_json(self) -> Dict:
         return {
             "node_type": "ValueNode",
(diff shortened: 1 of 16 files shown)


Sxnan:

python/feathub/feature_views/transforms/tests/test_over_window_transform.py
```diff
@@ -994,6 +994,44 @@ def test_over_window_on_join_field(self):
         self.assertListEqual(["name"], built_feature_view_2.keys)
         self.assertTrue(expected_result_df.equals(result_df))
 
+    def test_over_window_transform_feature_without_dtype(self):
```
lindong28, line 997: Should we update exiting tests to remove unnecessary specifications of `Feature#dtype`?

It might be useful to have tests to cover the scenario where dtype is different from the original type (e.g. int -> double and double -> int).

python/feathub/processors/flink/table_builder/flink_table_builder.py
```diff
@@ -578,14 +578,13 @@ def _evaluate_expression_transform(
...
     ) -> NativeFlinkTable:
-        result_type = to_flink_type(result_type)
-        return source_table.add_or_replace_columns(
-            native_flink_expr.call_sql(to_flink_sql_expr(transform.expr))
-            .cast(result_type)
-            .alias(result_field_name)
-        )
+        res_col = native_flink_expr.call_sql(to_flink_sql_expr(transform.expr))
+        if result_type is not None:
+            result_type = to_flink_type(result_type)
```
lindong28, line 585: With the implementation in this PR, we will leave the derivation of `Feature#dtype` to the underlying processor (e.g. Flink, Spark, Pandas). 

It means that FeatHub won't be able to tell what is the type of features emitted by its FeatureView. However, since FeatHub provides its own SDK and DSl for users to define schema, it should be able to derive and store the expected type of a FeatureView, just like what Spark/Flink does.

Also, would there be inconsistency between the dtypes derived by different processors?

Would it be better to derive FeatHub's Feature#dtype when we resolve a FeatureView and still ask the underlying processor to cast the feature to this dtype?

python/feathub/feature_views/feature.py
```diff
@@ -35,17 +35,19 @@ class Feature:
...
         input_features: Sequence[Feature] = (),
     ):
         """
         :param name: The name that uniquely identifies this feature in the
                      parent table.
-        :param dtype: The data type of this feature's values.
         :param transform: The logic used to derive this feature's values. If it is a
                           string, it represents a Feathub expression.
+        :param dtype: Optional. If it is not None, it is the data type of this
+                      feature's values. If is it None, the data type is derived by the
```
lindong28, line 49: `by` -> `from`

lindong28 (commented):

Sxnan:
@lindong28 Thanks for your review! I updated the PR to introduce type derivation. Could you please take another look?

python/feathub/feature_views/feature_view.py
```diff
@@ -192,3 +214,50 @@ def get_bounded_view(self) -> TableDescriptor:
...
+        else:
+            raise FeathubException(
+                f"Cannot derive feature data type of {type(transform)} feature"
+            )
+
+    def _derive_feature_dtype_from_feathub_expression(self, expr: str) -> DType:
+        ast = feathub_expr_parser.parse(expr)
+        variable_types = {}
+
+        for input_feature_name in ast.get_variables():
```
lindong28, line 245: Instead of using `TableDescriptor#get_feature` to get the dtype for variables used in the given expression, an alternative approach is to get the dtype for all features defined on `source` and this FeatureView, and re-use this method to determine the dtype for arbitrary expressions used in this FeatureView.

The latter approach seems simpler since the same mapping can be repeatedly re-used and we don't need to add the abstract method `ExprAST#get_variables`. We can add `TableDescriptor#get_output_fields(self)` to expose this information.

What do you think?

python/feathub/dsl/ast_type_derive_utils.py
```diff
@@ -0,0 +1,224 @@
...
+    CaseOp,
+    NullNode,
+)
+
+TYPE_PRECISION_RANK: List[DType] = [Float64, Float32, Int64, Int32]
+
+# A map from the Feathub built-in function name to a callable that is used to
+# evaluate the result data type of the function. The callable takes a list of
+# data types of arguments and returns the data type of the result.
+FUNCTION_RESULT_TYPE_EVALUATOR: Dict[str, Callable[[List[DType]], DType]] = {
```
lindong28, line 50: We probably need a dedicated class to define all FeatHub built-in functions similar to the existing `AggFunc` class. Then we can put the function signature definition there and apply this definition across all processors, and use dedicatedd IT tests for these functions similar to how we test FeatureView subclasses for all processors.

Maybe add it as TODO if it is not easy to do in this PR?

lindong28 (commented):

python/feathub/dsl/ast.py
```diff
@@ -29,17 +29,43 @@ def to_json(self) -> Dict:
...
+
     def __str__(self) -> str:
         return json.dumps(self.to_json(), indent=2, sort_keys=True)
 
 
-class BinaryOp(ExprAST):
-    def __init__(self, op_type: str, left_child: ExprAST, right_child: ExprAST) -> None:
-        super().__init__(node_type="BinaryOp")
-        self.op_type = op_type
+class AbstractUnaryOp(ExprAST, ABC):
```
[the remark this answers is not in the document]
Sxnan: I think we still need to subclass ABC here. Otherwise, the abstract methods have to be implemented in the `AbstractUnaryOp`.

python/feathub/dsl/ast.py
```diff
@@ -161,10 +197,9 @@ def to_json(self) -> Dict:
         }
 
 
-class GroupNode(ExprAST):
+class GroupNode(AbstractUnaryOp):
```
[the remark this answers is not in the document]
Sxnan: I think we should keep the `GroupNode`. For example, in `FlinkProcessor` and `SparkProcessor`, where the Feathub Expr is translated to Flink SQL Expr and Spark SQL Expr, they can benefit from the GroupNode. Otherwise, they must put parentheses around all the binary operators in Flink SQL Expr and Spark SQL Expr.

python/feathub/dsl/ast_type_derive_utils.py
```diff
@@ -0,0 +1,224 @@
...
+    "LOWER": lambda _: String,
+    "UNIX_TIMESTAMP": lambda _: Int64,
+}
+
```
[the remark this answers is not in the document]
Sxnan: Thanks for the suggestion. It is indeed more readable.

python/feathub/dsl/ast_type_derive_utils.py
Sxnan: I agree adding a class for all Feathub built-in functions is necessary. And I think it should account for both built-in functions and user-defined functions (#72, #83). I added a TODO and issue #86 to keep track of it.

Sxnan:

python/feathub/dsl/ast.py
```diff
@@ -218,6 +351,24 @@ def __init__(
...
+        )
+
+        if not isinstance(self.default, NullNode):
+            result_types.add(self.default.eval_dtype(variable_types))
+
+        if len(result_types) == 1:
+            return result_types.pop()
+
+        result_type = _get_higher_precision_type(*result_types)
+        if result_type is None:
```
lindong28, line 366: Do we need this check?

python/feathub/feature_views/feature_view.py
```diff
@@ -136,18 +150,17 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
...
-            raise RuntimeError(
-                f"Failed to find the feature '{feature_name}' in {self.to_json()}."
+        output_features = {}
+        if self.keep_source_fields:
+            output_features.update(
+                {f.name: f for f in self.get_resolved_source().get_output_features()}
             )
-        return cast(TableDescriptor, self.source).get_feature(feature_name)
+        output_features.update({f.name: f for f in self.get_resolved_features()})
+        return list(output_features.values())
```
lindong28, line 163: Would it be useful to make the order of the features in this list the same as the order of features outputted by the FeatureView, so that we can reuse this list to construct the FeatureView's schema?

python/feathub/feature_views/feature_view.py
```diff
@@ -192,3 +205,38 @@ def get_bounded_view(self) -> TableDescriptor:
         feature_view = deepcopy(self)
         feature_view.source = self.get_resolved_source().get_bounded_view()
         return feature_view
+
+    def _derive_feature_dtype(self, feature: Feature) -> Optional[DType]:
+        transform = feature.transform
+        if isinstance(transform, ExpressionTransform):
+            return self._derive_feature_dtype_from_feathub_expression(transform.expr)
+        elif isinstance(transform, OverWindowTransform):
```
lindong28, line 213: Use `isinstance(transform, OverWindowTransform) or isinstance(transform, SlidingWindowTransform)` here?

python/feathub/feature_views/feature_view.py
```diff
@@ -192,3 +205,38 @@ def get_bounded_view(self) -> TableDescriptor:
...
+                transform.expr
+            )
+            return transform.agg_func.get_result_type(expr_result_type)
+        elif isinstance(transform, SlidingWindowTransform):
+            expr_result_type = self._derive_feature_dtype_from_feathub_expression(
+                transform.expr
+            )
+            return transform.agg_func.get_result_type(expr_result_type)
+        elif isinstance(transform, JoinTransform):
+            # We don't need to derive data type of the feature to join. Its data type is
```
lindong28, line 224: We need to get the dtype of the feature to join to determine the dype of the output feature, right?

We should be able to know the dtype of every feature. If the program references a feature on another feature view in the feature registry, we should be able to retrieve the dtype of this feature from the feature registry. This is done in `_DerivedFeatureView#get_feature_from_feature_str`. In general, the dtype of the feature can be obtained when we resolve the feature view.

lindong28 (commented):
Thanks for the update! It is much better now. Left some comments below.

python/feathub/dsl/ast.py
```diff
@@ -29,17 +64,48 @@ def to_json(self) -> Dict:
...
+        self.op_type = op_type
+
+    def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
+        left_type = self.left_child.eval_dtype(variable_types)
+        right_type = self.right_child.eval_dtype(variable_types)
+        if left_type == right_type:
+            return left_type
+
+        res_type = _get_higher_precision_type(left_type, right_type)
+        if res_type is None:
```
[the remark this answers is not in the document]
Sxnan: I think by the signature of `eval_dtype`, it should return a non-None value.

However, even when dtypes of left and right children are non-None values, the `res_type` can be None. _get_higher_precision_type can return None if it sees a type other than Int and Float. For example, it can encounter this Feathub expr `'abc' * 'de'`

python/feathub/feature_views/feature_view.py
Sxnan: You are right that we should know the data type of the join feature. Actually, this method should not be called with the JoinTransform feature, as its data should not be None. So I think it is better to throw an exception here.

python/feathub/dsl/ast.py
Sxnan: I think we need this check for the same reason above.

python/feathub/dsl/ast.py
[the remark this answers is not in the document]
lindong28: Would it be simpler to move this check to `_get_higher_precision_type()`?

python/feathub/feature_views/feature_view.py
```diff
@@ -136,18 +150,22 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
...
-                f"Failed to find the feature '{feature_name}' in {self.to_json()}."
-            )
-        return cast(TableDescriptor, self.source).get_feature(feature_name)
+        source_features = self.get_resolved_source().get_output_features()
+        features = {
+            **{f.name: f for f in source_features},
+            **{f.name: f for f in self.get_resolved_features()},
+        }
+
+        output_feature_names = self.get_output_fields([f.name for f in source_features])
```
lindong28, line 163: Would it be simpler to remove `FeatureView#get_output_fields`?

Only OnDemandFeatureView needs `get_output_fields(self, source_fields: List[str])` because it does not have source information available when it is constructed. We can add `OnDemandFeatureView#get_output_features(self, source_fields: List[str])`.

Please feel free to create a TODO for this task if you prefer.

python/feathub/feature_views/feature_view.py
```diff
@@ -192,3 +210,33 @@ def get_bounded_view(self) -> TableDescriptor:
...
+            raise FeathubException("JoinTransform feature should have dtype set.")
+        else:
+            raise FeathubException(
+                f"Cannot derive feature data type of {type(transform)} feature"
+            )
+
+    def _derive_feature_dtype_from_feathub_expression(self, expr: str) -> DType:
+        ast = feathub_expr_parser.parse(expr)
+
+        variable_types = {
```
lindong28, line 235: Can we derive the `variable_types` once and re-use it to derive types for all Features in this FeatureView?

lindong28 (commented):
Thanks for the update. Left some minor comments below.

python/feathub/feature_views/feature_view.py
Sxnan: I added a TODO.

Sxnan:
@lindong28 Thanks for your review. I updated the PR and all comments should be addressed. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #75, opened 2023-02-14
FeatureView overwrites feature in source

Sxnan:

FeatureView overwrites the feature in the source if they have the same name when using FlinkProcessor

## Brief change log

- Update FlinkProcessor to overwrite the feature in the source if the feature in FeatureView has the same name.

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/feature_views/derived_feature_view.py
+++ b/python/feathub/feature_views/derived_feature_view.py
@@ -13,13 +13,16 @@
 # limitations under the License.
 
 from __future__ import annotations
+
 from typing import Union, Dict, Sequence, Optional
 
-from feathub.feature_views.transforms.join_transform import JoinTransform
-from feathub.table.table_descriptor import TableDescriptor
+from feathub.common.exceptions import FeathubException
 from feathub.feature_views.feature import Feature
 from feathub.feature_views.feature_view import FeatureView
+from feathub.feature_views.transforms.expression_transform import ExpressionTransform
+from feathub.feature_views.transforms.join_transform import JoinTransform
 from feathub.registries.registry import Registry
+from feathub.table.table_descriptor import TableDescriptor
 
 
 class DerivedFeatureView(FeatureView):
@@ -51,8 +54,10 @@ class DerivedFeatureView(FeatureView):
                          computed by ExpressionTransform or OverWindowTransform, its
                          expression can only depend on the features specified earlier in
                          this list and the features in the source table.
-        :param keep_source_fields: True iff all fields in the source table should be
-                                   included in this table.
+        :param keep_source_fields: True iff all features in the source table should be
+                                   included in this table. The feature in the source
+                                   will be overwritten by the feature in this feature
+                                   view if they have the same name.
         :param filter_expr: Optional. If it is not None, it represents a Feathub
                             expression which evaluates to a boolean value. The filter
                             expression is evaluated after other transformations in the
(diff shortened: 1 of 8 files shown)


Sxnan:

python/feathub/feature_views/transforms/tests/test_expression_transform.py
```diff
@@ -268,3 +268,74 @@ def _test_case(self, upper_name_feature: Feature):
...
+        expected_result_df["cost"] = expected_result_df.apply(
+            lambda row: row["cost"] + 1, axis=1
+        )
+        expected_result_df = expected_result_df.sort_values(by=["time"]).reset_index(
+            drop=True
+        )
+
+        self.assertTrue(expected_result_df.equals(result_df))
+
+    def test_refer_to_source_feature(self):
```
lindong28, line 303: It seems that this test includes the FeatHub program used in `test_duplicate_feature_name_with_source`. Would it be simpler to check the result of `feature_view_1` in this test and remove the other test?

The name `test_duplicate_feature_name_with_source` is probably better.

python/feathub/feature_views/tests/test_feature_view.py
```diff
@@ -78,7 +78,7 @@ def test_features(self):
         )
         built_feature_view = self.registry.build_features([feature_view_2])[0]
 
-        self.assertEqual(feature_1, built_feature_view.get_feature("feature_1"))
+        self.assertIsNotNone(built_feature_view.get_feature("feature_1"))
```
lindong28, line 81: If it is not equal `feature_1`, what would be the expected value? Can we specify the expected value here?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -47,7 +50,10 @@ def __init__(
                          feature is a string, it should be either in the format
                          {table_name}.{feature_name}, which refers to a feature in the
                          table with the given name, or in the format {feature_name},
-                         which refers to a feature in the source table. For any feature
+                         which refers to a feature in the source table. If
```
lindong28, line 53: Even if we don't add this paragraph in the doc, the existing doc already says that we will compute these `features`. So the doc seems redundant here, right?

Should we update the doc for `keep_source_fields`?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -98,7 +104,8 @@ def build(
                 table_desc = registry.get_features(name=join_table_name)
                 join_feature = table_desc.get_feature(feature_name=join_feature_name)
                 if source.name == join_table_name:
-                    feature = join_feature
+                    feature = deepcopy(join_feature)
```
lindong28, line 107: Can you explain why we should copy the `input_features` from the `join_features`?

lindong28 (commented):

Sxnan:
@lindong28 Thanks for your review. I updated the PR. Could you please take another look?

python/feathub/feature_views/derived_feature_view.py
lindong28: Is this comment addressed?

Sxnan: My bad. I misunderstood the comment and addressed the wrong problem. I think it is resolved now.

Sxnan:
@lindong28 I updated the PR to address the last comment. Could you please take another look?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -125,6 +107,44 @@ def build(
...
+    ) -> Feature:
+        parts = feature_str.split(".")
+        if len(parts) == 1:
+            source_feature = source.get_feature(parts[0])
+            feature = Feature(
+                name=source_feature.name,
+                dtype=source_feature.dtype,
+                transform=ExpressionTransform(source_feature.name),
+                keys=source_feature.keys,
+                input_features=(),
```
lindong28, line 122: nits: it seems redundant to specify the default value for `input_features`.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -209,7 +209,14 @@ def build(
...
         for feature in self.features:
             if isinstance(feature, str):
-                feature = source.get_feature(feature_name=feature)
+                source_feature = source.get_feature(feature)
+                feature = Feature(
+                    name=source_feature.name,
+                    dtype=source_feature.dtype,
+                    transform=ExpressionTransform(source_feature.name),
+                    keys=source_feature.keys,
+                    input_features=(),
```
lindong28, line 218: nits: it seems redundant to specify the default value for `input_features`.

lindong28 (commented):

Sxnan:
@lindong28 Thanks for your review. I updated the PR. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #77, opened 2023-02-19
Support adding late data to sliding window in FlinkProcessor

yunfengzhou-hub:

This pull request adds support for adding late data to sliding windows in FlinkProcessor.

For example, suppose there is a SlidingWindowTransform whose window_size is 15min and step_size is 1min, and the source's max_out_of_orderness is 0. If a row of features, whose timestamp is 8:59, arrives at the Feathub job when the watermark is at 9:01, this row will be dropped and not considered in any window. With this pull request, this row will not be considered in the window 8:45 ~ 9:00, but Feathub would be able to add it to all following windows that it belongs to, from 8:46~9:01 to 8:59~9:14.

## Brief change log

- Uses Flink DataStream API's windowing operation instead of SQL API's to expose the interface to access late data.
- Connects late data stream to SlidingWindowKeyedProcessFunction.

  - The public API: yes

## Documentation


java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
```diff
@@ -267,11 +278,75 @@ fieldName, getDataType(resolvedSchema, fieldName)))
...
+        }
+
+        @Override
+        public void apply(
+                Row keys, TimeWindow window, Iterable<Row> iterable, Collector<Row> collector) {
+            Row acc = Row.withNames();
+            for (AggregationFieldsDescriptor.AggregationFieldDescriptor fieldDescriptor :
+                    fieldsDescriptor.getAggFieldDescriptors()) {
+                acc.setField(
+                        fieldDescriptor.outFieldName, fieldDescriptor.aggFunc.createAccumulator());
```
lindong28, line 327: It seems weird to output accumulator instance instead of accumulator result. And it might add unexpected serialization overhead since we probably will not add custom serializer for accumulators.

Also note that the existing `RawDataAccumulator` keeps the entire history of records instead of merging them into one record before this accumulator instance is outputted here. This would effectively disable the pre-aggregation optimization used before this PR and considerably increase the memory usage.

Instead of putting accumulator instances in the output row, would it be more intuitive to output concrete values (e.g. accumulated results) similar to the approach prior to this PR?

lindong28 (commented):

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
```diff
@@ -61,73 +71,86 @@ public class SlidingWindowUtils {
...
+        final OutputTag<Row> outputTag = new OutputTag<Row>("late-data-side-output") {};
+
+        TimeZone timeZone = TimeZone.getTimeZone(tEnv.getConfig().getLocalTimeZone());
+
+        SingleOutputStreamOperator<Row> outputStream =
+                stream.keyBy(new RowKeySelector(keyFieldNames))
+                        .window(
+                                TumblingEventTimeWindows.of(
+                                        Time.milliseconds(stepSizeMs),
+                                        getModdedOffset(stepSizeMs, -timeZone.getRawOffset())))
```
Sxnan, line 104: We are not handling daylight saving. Should we add a TODO as a reminder?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedCoProcessFunction.java
```diff
@@ -189,15 +232,38 @@ public void onTimer(
...
+                List<Row> lateDatas = state.getLateDatas();
+                for (int i = 0; i < lateDatas.size(); i++) {
+                    if (!state.isLateDataReferencedBy(i, aggFieldIdx)) {
+                        continue;
+                    }
+
+                    Row lateData = lateDatas.get(i);
+                    long lateDataTimestamp =
+                            ((Instant) lateData.getFieldAs(rowTimeFieldName)).toEpochMilli();
+                    if (lateDataTimestamp > timestamp - descriptor.windowSizeMs + 1) {
```
Sxnan, line 254: Could you double-check whether this should be `>` or `>=`?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, Flink cannot handle cases when abs(offset) > abs(windowSize). Thus `getModdedOffset()` is necessary.

yunfengzhou-hub: According to offline discussion, Feathub may not need to handle daylight saving. We may assume each day have 24 hours.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedCoProcessFunction.java
```diff
@@ -150,12 +153,52 @@ public void processElement(
...
+                            - 1
+                            + stepSizeMs;
+            if (state.maxTriggeredTimer.value() >= expireTime) {
+                state.dereferenceLateDataByIndex(lateDataIndex, aggFieldIdx);
+                continue;
+            }
+
+            Object accumulator = accumulatorStates.getField(aggFieldIdx);
+            Object value = row.getField(descriptor.inFieldName);
+            descriptor.aggFunc.add(accumulator, value, timestamp);
```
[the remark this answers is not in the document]
yunfengzhou-hub: Simplifying the code here might introduce some performance overhead. For example, now we does the following steps:

1. read accumulator value
2. add some values to accumulator
3. retract some values from accumulator
4. save accumulator value

After we simplify them into functions:

1. invoke add_value() function
    a. read accumulator value
    a. add some values to accumulator
    a. save accumulator value
2. invoke retract_value() function
    a. read accumulator value
    a. retract some values from accumulator
    a. save accumulator value

Compared with the previous process, the simplified process read/write the accumulator value from flink state twice. We have not tested against this behavior, not knowing how much overhead it might bring to the performance.

yunfengzhou-hub:
Hi @lindong28, I have fixed previous comments and rebased the commit onto the latest master code. Could you please take a look at this PR?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -337,19 +394,56 @@ public static SlidingWindowState create(
...
+                    }
+                    timestampList.update(timestamps);
+                }
+
+                List<Integer> leftIdxList = getLeftTimestampIdx();
+                for (AggregationFieldsDescriptor.AggregationFieldDescriptor aggFieldDescriptor :
+                        aggregationFieldsDescriptor.getAggFieldDescriptors()) {
+                    int aggFieldIdx =
+                            aggregationFieldsDescriptor.getAggFieldIdx(aggFieldDescriptor);
+                    if (affectedAggFieldIndices != null
```
Sxnan, line 440: It appears that the affectedAggFieldIndices variable is used to determine whether the left index of an aggregator should be incremented by 1 or not. Instead of using this variable, we can check if the current left index of the aggregator is greater than or equal to the index of the added row. This way, we may be able to eliminate the need for the affectedAggFieldIndices variable.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -168,16 +209,18 @@ public void onTimer(
...
         final List<Integer> leftIdxList = state.getLeftTimestampIdx();
 
         for (AggregationFieldsDescriptor.AggregationFieldDescriptor descriptor :
                 aggregationFieldsDescriptor.getAggFieldDescriptors()) {
             final int aggFieldIdx = aggregationFieldsDescriptor.getAggFieldIdx(descriptor);
             Object accumulatorState = accumulatorStates.getField(aggFieldIdx);
-            if (rowToAdd != null) {
-                descriptor.aggFunc.merge(accumulatorState, rowToAdd.getField(descriptor.fieldName));
+            if (state.timestampToRow.contains(timestamp)) {
+                for (Row rowToAdd : state.timestampToRow.get(timestamp)) {
```
Sxnan, line 220: I believe that we can get the rows to add once outside of the loop to avoid accessing the state multiple times.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
yunfengzhou-hub: If the left index of an aggregation field is equal to the added row's index, we are unsure whether the left idx should increment by 1. `affectedAggFieldIndices` works in this situation.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -238,7 +283,7 @@ private static class SlidingWindowState {
         private final AggregationFieldsDescriptor aggregationFieldsDescriptor;
 
         /** This MapState maps from row timestamp to the row. */
-        private final MapState<Long, Row> timestampToRow;
+        private final MapState<Long, List<Row>> timestampToRow;
```
Sxnan, line 286: Could you also rename the variable to `timestampToRows` and also the docs to reflect the change?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
```diff
@@ -108,31 +118,68 @@ public static DataStream<Row> applySlidingWindowPreAggregationProcess(
...
+                                        offset))
+                        .returns(
+                                Types.ROW_NAMED(
+                                        fieldNames.toArray(new String[0]),
+                                        resultFieldTypes.toArray(new TypeInformation[0])));
+
+        return resultStream.union(lateDataStream);
+    }
+
+    public static AggregationFieldsDescriptor getAggregationFieldsDescriptorsWithLimit(
```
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, I have created #97 to record this as a TODO in future.

Sxnan:
@yunfengzhou-hub Thanks for your update! LGTM. @lindong28 Could you help review this PR?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -122,15 +118,62 @@ public void processElement(
...
+                if (timeToRetract <= maxTriggeredTimer) {
+                    continue;
+                }
+
+                affectedFieldIndices.add(aggFieldIdx);
+
+                Object accumulatorState = accumulatorStates.getField(aggFieldIdx);
+                Object preAggResult = row.getField(descriptor.fieldName);
+                descriptor.aggFunc.merge(accumulatorState, preAggResult);
+                accumulatorStates.setField(aggFieldIdx, accumulatorState);
```
lindong28, line 161: Do we still need this?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -122,15 +118,62 @@ public void processElement(
...
+                if (skipSameWindowOutput) {
+                    ctx.timerService().registerEventTimeTimer(timeToRetract);
+                }
+            }
+
+            state.addRow(rowTime, row, affectedFieldIndices);
+            state.updateAccumulatorStates(accumulatorStates);
+
+            if (skipSameWindowOutput) {
+                ctx.timerService().registerEventTimeTimer(maxTriggeredTimer + stepSizeMs);
```
lindong28, line 172: Why do we need this?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -188,9 +232,10 @@ public void onTimer(
                 if (timestamp - descriptor.windowSizeMs < rowTime) {
                     break;
                 }
-                Row curRow = state.timestampToRow.get(rowTime);
-                descriptor.aggFunc.retractAccumulator(
-                        accumulatorState, curRow.getField(descriptor.fieldName));
+                for (Row curRow : state.timestampToRows.get(rowTime)) {
```
lindong28, line 235: It would be better to rename `curRow` to be consistent with `rowToAdd` above.

Maybe rename both variables as `row`.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -333,23 +391,60 @@ public static SlidingWindowState create(
...
+                timestampToRows.put(timestamp, rows);
+            } else {
+                timestampToRows.put(timestamp, new ArrayList<>(Collections.singletonList(row)));
+
+                List<Long> timestamps = getTimestampList();
+
+                if (timestamps.isEmpty() || timestamps.get(timestamps.size() - 1) < timestamp) {
+                    timestampList.add(timestamp);
+                } else {
+                    for (int i = 0; i < timestamps.size(); i++) {
```
lindong28, line 426: Let's add a TODO to use skip list for timestamps so that late data can be handled in O(n) time.

lindong28 (commented):

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/aggregation/FirstLastValueAggFuncWithoutRetract.java
```diff
@@ -69,13 +78,29 @@ public void merge(FirstLastValueAccumulator<T> target, FirstLastValueAccumulator
...
         throw new UnsupportedOperationException();
     }
+
+    @Override
+    public DataType getResultDatatype() {
+        return inDataType;
+    }
+
+    @Override
+    public AbstractRawDataAggFunc.RawDataAccumulator<T> createAccumulator() {
```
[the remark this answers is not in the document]
yunfengzhou-hub: We need to make sure that `FirstLastValueAggFuncWithoutRetract` and `FirstLastValueAggFunc` use the same accumulator type, so that SlidingWindowKeyedProcessFunction can handle the accumulators passed from the pre-agg process.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/SlidingWindowUtils.java
```diff
@@ -77,6 +81,12 @@ public static DataStream<Row> applySlidingWindowPreAggregationProcess(
             table = table.filter(Expressions.callSql(windowDescriptor.filterExpr));
         }
 
+        if (windowDescriptor.limit != null) {
+            aggDescriptors =
+                    getAggregationFieldsDescriptorsWithLimit(
```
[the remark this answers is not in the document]
yunfengzhou-hub: Currently for either Java or Python code, the concept of "limit" only exists on `WindowDescriptor`, not on `AggregationFieldDescriptor`. According to offline discussions with @Sxnan, a complete optimization of this problem might be as follows:

1. remove `limit` parameter from `WindowDescriptor`.
2. add `limit` parameter to `AggregationFieldDescriptor`, and get `aggFuncWithoutRetract` during instantiation as described in the comment.
3. handle aggregation fields with different `limit` values together in one sliding/over window evaluation process, instead of handling them separately and joining them afterward.

Given that it might be improper only to apply the second step above in this PR, and the complete steps seem to big to be put in this PR. I have created #97 to track this problem and maybe we can fix this in the next PR.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/AggregationFieldsDescriptor.java
```diff
@@ -114,17 +113,27 @@ public AggregationFieldDescriptor(
...
-                            AggFuncUtils.getAggFunc(aggFuncName, inDataType, true);
-
-            this.aggFuncWithoutRetract =
+                            AggFuncUtils.getAggFunc(aggFuncName, inDataType, true),
                     (AggFunc<Object, ?, Object>)
-                            AggFuncUtils.getAggFunc(aggFuncName, inDataType, false);
+                            AggFuncUtils.getAggFunc(aggFuncName, inDataType, false));
+        }
+
+        public AggregationFieldDescriptor(
```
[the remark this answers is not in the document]
yunfengzhou-hub: The solution of the comment depends on that of the next comment. Please refer to the next comment for discussion thread.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
yunfengzhou-hub: We need to make sure that after handling late data, processfunction can output the updated agg result right in the first time, instead of waiting for other potential non-late data to trigger the output.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/processfunction/SlidingWindowKeyedProcessFunction.java
```diff
@@ -333,23 +391,60 @@ public static SlidingWindowState create(
...
+                }
+
+                List<Integer> leftIdxList = getLeftTimestampIdx();
+                for (AggregationFieldsDescriptor.AggregationFieldDescriptor aggFieldDescriptor :
+                        aggregationFieldsDescriptor.getAggFieldDescriptors()) {
+                    int aggFieldIdx =
+                            aggregationFieldsDescriptor.getAggFieldIdx(aggFieldDescriptor);
+                    if (affectedAggFieldIndices != null
+                            && !affectedAggFieldIndices.contains(aggFieldIdx)) {
+                        leftIdxList.set(aggFieldIdx, leftIdxList.get(aggFieldIdx) + 1);
```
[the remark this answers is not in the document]
yunfengzhou-hub: - After inserting the timestamp and row into states, we need to update left idx list so that left indices can correctly reflect the rows to retract.
- For each aggregation field, we need to compare `row time` and `maxTriggeredTimer - windowSizeMs` to judge whether `leftIdx` should increment by 1 or not. We cannot directly compare `leftIdx` and the row's insertion index(`insertIdx`), because when `leftIdx == insertIdx`, both cases are possible.
- The comparison of `row time` and `maxTriggeredTimer - windowSizeMs` has been performed in `processElement`, so I choose to record that result in `affectedAggFieldIndices` variable and reuse the result here.

yunfengzhou-hub:
Thanks for the comments @lindong28. I have updated the PR according to the comments and offline discussions.

lindong28:
Thanks for the update. LGTM.

PR #91, opened 2023-03-21
Introduce MySQLSource and MySQLSink

Sxnan:

Introduce MySQLSource and MySQLSink so that Feathub can write features to MySQL table and use MySQL as online store.

## Brief change log

- Introduce MySQLSink
- Introduce MySQLSource and MySQLClient to use MySQL as online store

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? Python Docstring and Doc

### DIFF
--- a/docs/feature_tables/mysql.md
+++ b/docs/feature_tables/mysql.md
@@ -1,0 +1,78 @@
+# MySQL
+
+Feathub provides `MySQLSource` to read data from a MySQL table and `MySQLSink` to 
+materialize feature view to a MySQL table. Currently, `MySQLSource` can only be used as 
+an online store source.
+
+## Supported Processors and Usages
+
+- Flink: Online Store Source, Append Sink, Upsert Sink
+
+## Example
+
+Here are the examples of using `MySQLSource` and `MySQLSink`:
+
+### Use as Append/Upsert Sink
+
+If the feature view to be materialized has keys, the MySQLSink is in upsert mode 
+otherwise it is in append mode.
+
+```python
+feature_view = DerivedFeatureView(...)
+
+sink = MySQLSink(
+    database="database",
+    table="table",
+    host="localhost",
+    username="user",
+    password="password",
+)
+
+featub_client.materialize_features(
+    features=feature_view,
+    sink=sink,
+    allow_overwrite=True,
+).wait(30000)
+```
+
+### Use as Online Store Source
+
+```python
+feature_table_schema = (
+    Schema.new_builder()
+    .column("id", types.Int64)
+    .column("feature_1", types.Int64)
+    .column("ts", types.String)
+    .build()
+)
+
+source = MySQLSource(
+    name="feature_table",
+    database="database",
+    table="table",
+    schema=feature_table_schema,
+    host="localhost",
+    username="user",
+    password="password",
+    keys=["id"],
+    timestamp_field="ts",
+)
+
+on_demand_feature_view = OnDemandFeatureView(
+    name="on_demand_feature_view",
+    features=[
+        "feature_table.feature_1",
+    ],
+    keep_source_fields=True,
+    request_schema=Schema.new_builder().column("id", types.Int64).build(),
+)
+
+feathub_client.build_features([source, on_demand_feature_view])
+
+request_df = pd.DataFrame(np.array([[2], [3]]), columns=["id"])
+
+online_features = feathub_client.get_online_features(
+    request_df=request_df,
+    feature_view=on_demand_feature_view,
+)
+```
--- a/python/feathub/feature_tables/tests/test_mysql_source_sink.py
+++ b/python/feathub/feature_tables/tests/test_mysql_source_sink.py
@@ -1,0 +1,187 @@
+#  Copyright 2022 The Feathub Authors
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+from abc import ABC
+from datetime import datetime
+from typing import Optional
+
+import pandas as pd
+import sqlalchemy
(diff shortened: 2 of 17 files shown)


python/feathub/tests/mysql_test_base.py
```diff
@@ -0,0 +1,47 @@
...
+_mysql_container: Optional[MySqlContainer] = None
+
+
+class MySQLITTestBase(unittest.TestCase):
+    """
+    Base class for MySQL IT test. It reuses a global mysql container for all the MySQL
+    IT test cases.
+    """
+
+    mysql_container: Optional[MySqlContainer] = None
```
lindong28, line 28: Would it be simpler to do the following so that we don't a dedicated class for each container (e.g. redis, mysql)? 

- Add the static variable `FeathubITTestBase#mysql_container`.
- Add static method `FeathubITTestBase#setup_mysql_container` and `FeathubITTestBase#teardown_mysql_container`
- Let `MySQLSourceSinkITTest#setUpClass` invoke `setup_mysql_container`. And let `MySQLSourceSinkITTest#teardownUpClass` invoke `teardown_mysql_container`.

Sxnan:
@lindong28 Thanks for your review! I updated the PR and added documentation about the MySQL feature table. Could you please take another look?

python/feathub/online_stores/tests/test_mysql_online_store.py
```diff
@@ -0,0 +1,188 @@
...
+            request_schema=Schema.new_builder().column("id", types.Int64).build(),
+        )
+
+    def tearDown(self) -> None:
+        super().tearDown()
+        engine = sqlalchemy.create_engine(self.mysql_container.get_connection_url())
+        with engine.begin() as conn:
+            conn.execute(sqlalchemy.text(f"DROP TABLE IF EXISTS {self.table_name}"))
+
+    def test_get_online_features(self):
```
lindong28, line 109: We implemented mysql client rather than mysql online store in FeatHub. This test is actually testing `MySQLSource`.

Would it be more intuitive to move this test to `MySQLSourceSinkITTest`?

python/feathub/feature_tables/tests/test_mysql_source_sink.py
```diff
@@ -0,0 +1,180 @@
...
+    def _get_mysql_sink(self, table_name):
+        return MySQLSink(
+            host=self.mysql_container.get_container_host_ip(),
+            port=int(
+                self.mysql_container.get_exposed_port(
+                    self.mysql_container.port_to_expose
+                )
+            ),
+            database=self.mysql_container.MYSQL_DATABASE,
+            username=self.mysql_container.MYSQL_USER,
```
lindong28, line 177: Can we make the order of parameters consistent with their order in the function declaration?

python/feathub/online_stores/tests/test_mysql_online_store.py
```diff
@@ -0,0 +1,188 @@
...
+            pd.DataFrame([[2, 2], [3, 3]], columns=["id", "val"])
+            .sort_values(by=["id"])
+            .reset_index(drop=True)
+        )
+
+        assert_frame_equal(expected_result, online_features)
+
+    def test_more_input_column_than_keys(self):
+        source = MySQLSource(
+            name="feature_table",
```
lindong28, line 149: It is inconsistent that we use the string `feature_table` and the variable `self.table_name` interchangeably in this file.

python/feathub/online_stores/tests/test_mysql_online_store.py
```diff
@@ -0,0 +1,188 @@
...
+            ),
+            database=self.mysql_container.MYSQL_DATABASE,
+            username=self.mysql_container.MYSQL_USER,
+            password=self.mysql_container.MYSQL_PASSWORD,
+            keys=["id"],
+            timestamp_field="ts",
+            table=self.table_name,
+        )
+
+        self.client.build_features([source, self.on_demand_feature_view])
```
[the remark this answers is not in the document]
Sxnan: I made the definition of the `source` and `on_demand_feature_view` next to each other. It should be easier to see the relationship between them

Sxnan:

docs/feature_tables/mysql.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Supported Processors and Usages
+
+- Flink: Online Store Source, Append Sink, Upsert Sink
+
+## Example
+
+Here are the examples of using `MySQLSource` and `MySQLSink`:
+
+### Use as Sink
```
lindong28, line 15: Would it be useful to mention Append/Upsert to match the content in `Supported Processors and Usages`?

python/feathub/online_stores/mysql_client.py
```diff
@@ -0,0 +1,113 @@
...
+from feathub.table.schema import Schema
+
+
+class MySQLClient(OnlineStoreClient):
+    def __init__(
+        self,
+        schema: Schema,
+        host: str,
+        port: int,
+        database: str,
```
lindong28, line 29: Update the parameter order to be consistent with `MySQLSource`?

python/feathub/feature_tables/tests/test_mysql_source_sink.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+        source = self.create_file_source(
+            input_data,
+            keys=["id"],
+            schema=schema,
+            timestamp_field="ts",
+            timestamp_format="%Y-%m-%d %H:%M:%S",
+        )
+        sink = self._get_mysql_sink(table_name)
+        self.client.materialize_features(
```
lindong28, line 79: It seems that the logic here pretty much duplicate the logic in `_initialize_mysql_table()`. Should we re-use _initialize_mysql_table()?

It might be reasonable to remove this test if the logic in this test is a subset of the logic in `test_get_online_features()`.

lindong28 (commented):

docs/feature_tables/mysql.md
Sxnan: The sink can be in append mode or upsert mode in the given example, depending on whether the given feature view has keys. I updated the docs accordingly.

python/feathub/feature_tables/tests/test_mysql_source_sink.py
Sxnan: I update the test of MySQLSink to reuse the `_initialize_mysql_table` method. But I think we should keep both tests, as they test different aspects and use cases.

Sxnan:
@lindong28 Thanks for your review! I updated the PR. Could you take another look?

lindong28:
Thanks for the update! LGTM.

PR #93, opened 2023-03-24
LocalProcessor supports SlidingWindowTransform

Sxnan:

Update LocalProcessor to support SlidingWindowTransform.

## Brief change log
- Update LocalProcessor to support SlidingWindowTransform

  - The public API: no

## Documentation


Sxnan:

python/feathub/common/utils.py
```diff
@@ -88,6 +88,24 @@ def append_unix_time_column(
     return unix_time_column
 
 
+def append_epoch_millis_column(
```
lindong28, line 91: Can we unify this method with the existing method `append_unix_time_column`?

It seems a bit weird to put this method (and also `append_unix_time_column`) in this file given that this method is expected to be used only by LocalProcessor. Can you help refactor it similar to the file pattern used by FlinkProcessor and SparkProcessor?

python/feathub/common/utils.py
```diff
@@ -88,6 +88,24 @@ def append_unix_time_column(
...
+    df: pd.DataFrame, timestamp_field: str, timestamp_format: str, tz: tzinfo
+) -> str:
+    epoch_millis_column = "_epoch_millis_time"
```
lindong28, line 94: Would it be more consistent with other processors to use `EVENT_TIME_ATTRIBUTE_NAME`?

python/feathub/processors/local/ast_evaluator/local_func_evaluator.py
```diff
@@ -12,11 +12,31 @@
...
+    format: str = "%Y-%m-%d %H:%M:%S",
+    tz: tzinfo = timezone.utc,
+) -> Optional[int]:
+    if time is None:
+        return None
+    try:
+        return int(to_unix_timestamp(time, format=format, tz=tz))
+    except ValueError as v:
+        # Strip off the unconverted part of the given time in str. The behavior is
+        # consistent with FlinkProcessor and SparkProcessor
```
lindong28, line 31: Can you explain where this logic is implemented for FlinkProcessor and SparkProcessor?

And if we want to enforce this logic across all processors, would it be simpler to do it in `to_unix_timestamp`?

python/feathub/processors/local/sliding_window_utils.py
```diff
@@ -0,0 +1,350 @@
...
+from feathub.feature_views.transforms.agg_func import AggFunc
+from feathub.feature_views.transforms.sliding_window_transform import (
+    SlidingWindowTransform,
+)
+from feathub.processors.local.aggregation_utils import AGG_FUNCTIONS
+from feathub.processors.local.ast_evaluator.local_ast_evaluator import LocalAstEvaluator
+from feathub.processors.local.type_utils import cast_dataframe_dtype
+
+
+class AggregationFieldDescriptor:
```
lindong28, line 35: Can we unify classes like this across all processors? Maybe add a TODO for future work.

lindong28 (commented):

python/feathub/processors/local/ast_evaluator/local_func_evaluator.py
Sxnan: After running some tests, It turns out that this behavior is the same as FlinkProcessor but different from SparkProcessor. In FlinkProcessor the unconverted part is dropped, while the Spark job will fail if there are unconverted parts.

I think Feathub should just pick one of them. And we need to implement the behavior for the other processor using UDF. I think the behavior in Flink is more robust. What do you think?

python/feathub/processors/local/type_utils.py
```diff
@@ -0,0 +1,49 @@
...
+    Cast the given Pandas Series to the given dtype.
+    """
+    if series.dtype == target_dtype:
+        return series
+
+    if target_dtype == str:
+        # Handle str type specially to avoid casting None value to 'None' str
+        return series.apply(lambda x: None if x is None else str(x))
+
+    return series.astype(target_dtype, errors="ignore")
```
lindong28, line 30: Can you explain what errors are ignored?

And if the goal is to keep the value as None if it is None, would it be simpler to do this:

```
return series.apply(lambda x: None if x is None else target_dtype(x))
```

python/feathub/processors/local/local_processor.py
```diff
@@ -478,3 +485,121 @@ def _evaluate_python_udf_transform(
...
+            window_descriptor=sliding_window_descriptor,
+            agg_descriptors=agg_field_descriptors,
+            tz=self.timezone,
+            parser=self.parser,
+            ast_evaluator=self.ast_evaluator,
+        )
+
+        for feature in per_row_transform_features_following_first_sliding_feature:
+            if isinstance(feature.transform, ExpressionTransform):
+                if feature.transform.expr == "CURRENT_EVENT_TIME()":
```
lindong28, line 575: Why do we need this `if` statement?

python/feathub/processors/local/sliding_window_utils.py
```diff
@@ -0,0 +1,353 @@
...
+        feature_view.get_resolved_source().timestamp_format,
+        tz,
+    )
+    df_copy[EVENT_TIME_ATTRIBUTE_NAME] = (
+        df_copy[EVENT_TIME_ATTRIBUTE_NAME] * 1000
+    ).astype(np.int)
+
+    if len(window_descriptor.group_by_keys) > 0:
+        agg_df = df_copy.groupby(by=window_descriptor.group_by_keys).apply(
+            _get_sliding_window_func(
```
lindong28, line 142: Would it be a bit simpler to do this:

```
agg_df = df_copy.groupby(by=window_descriptor.group_by_keys).apply(
    lambda df: _sliding_window_func(
        df=df,
        sliding_window_descriptor=window_descriptor,
        agg_field_descriptors=agg_descriptors,
        tz=tz,
        parser=parser,
        ast_evaluator=ast_evaluator,
    )
)
```

python/feathub/processors/local/sliding_window_utils.py
```diff
@@ -0,0 +1,353 @@
...
+                sliding_window_descriptor=window_descriptor,
+                agg_field_descriptors=agg_descriptors,
+                tz=tz,
+                parser=parser,
+                ast_evaluator=ast_evaluator,
+            )
+        )
+    else:
+        agg_df = df_copy.sort_values(by=[EVENT_TIME_ATTRIBUTE_NAME])
```
lindong28, line 151: Do we also need to sort df by EVENT_TIME_ATTRIBUTE_NAME when `len(window_descriptor.group_by_keys) > 0`?

python/feathub/processors/local/sliding_window_utils.py
```diff
@@ -0,0 +1,353 @@
...
+    ast_evaluator: LocalAstEvaluator,
+) -> Optional[pd.DataFrame]:
+
+    df_copy = df.copy()
+    if df_copy.shape[0] <= 0:
+        return None
+
+    df_copy.sort_values(by=[EVENT_TIME_ATTRIBUTE_NAME], inplace=True)
+    df_copy[EVENT_TIME_ATTRIBUTE_NAME] = df_copy[EVENT_TIME_ATTRIBUTE_NAME].transform(
+        lambda t: _to_local_timestamp_mills(t, tz)
```
lindong28, line 220: Can you explain why we need this conversion?

python/feathub/processors/local/ast_evaluator/local_func_evaluator.py
lindong28: Since we want to consistently dropped the uncovered part, would it be simpler to implement this logic in `feathub.common.utils.to_unix_timestamp`?

lindong28 (commented):

python/feathub/processors/local/ast_evaluator/local_func_evaluator.py
Sxnan: Thanks for the explanation. I also find it to be hacky to use the error message format. I agree we should go with the Spark behavior. And I added a TODO to FlinkProcessor to add a UDF to make the behavior consistent.

python/feathub/processors/local/type_utils.py
Sxnan: It ignores the error when casting a Series with nan value to integer. I updated the code to handle that case explicitly and the removed the `ignore`.

python/feathub/processors/local/local_processor.py
Sxnan: The event time is computed in the `evaluate_sliding_window` method. This is a temporary solution to ignore the `CURRENT_EVENT_TIME` function. 

I added a TODO to properly handle the `CURRNET_EVENT_TIME` in the future.

python/feathub/processors/local/sliding_window_utils.py
Sxnan: We need to convert the unix time to the local timestamp mills that describe the number of milliseconds since 1970-01-01 00:00:00 at the local timezone. We assign row based on the local timestamp millis instead of unix time so that the windows are aligned with 1970-01-01 00:00:00 at the current time zone.

python/feathub/processors/local/time_utils.py
```diff
@@ -0,0 +1,46 @@
+#  Copyright 2022 The Feathub Authors
```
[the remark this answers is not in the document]
Sxnan: I think you are referring to `feathub.processors.spark.dataframe_builder.source_sink_utils_common`. And I renamed it to `time_utils.py`.

Sxnan:
@lindong28 Thanks for the review. I updated the PR and all comments should be addressed. Could you please take another look?

lindong28:
Thanks for the update. LGTM.

PR #94, opened 2023-03-24
Support reordering of output features in feature view

jiangxin369:

Support reordering of output features in feature view. The order of output fields would follow these principles:
* If `keep_source_fields` is not set, select all derived fields in the order that the user assigned.
* If `keep_source_fields` is set, select all source fields that are not overridden first, then select derived fields.

## Brief change log

  - Replace the current reordering logic with the above principles.

  - The public API: (no)

## Documentation

### DIFF
--- a/python/feathub/feature_views/feature_view.py
+++ b/python/feathub/feature_views/feature_view.py
@@ -121,10 +121,14 @@ class FeatureView(TableDescriptor, ABC):
         """
         Returns the names of fields of this table descriptor. This method should be
         called after the FeatureView is resolved, otherwise exception will be raised.
-        The output fields include:
-        - All fields in the source_fields if keep_source_fields is True.
-        - The timestamp field if it is not None.
-        - All features and features' keys.
+        The output fields and orders follows the below principle:
+
+        - If keep_source_fields is True, outputs all source fields that are not
+          overwritten by features in the order that they appear in the source fields.
+        - If keep_source_fields is False, outputs the timestamp field and features'
+          keys that are not overwritten by features in the order that they appear in the
+          source fields.
+        - Outputs all features in the order specified by users.
 
         :param source_fields: The names of fields of the source table.
         :return: The names of fields of this table descriptor.
@@ -133,25 +137,21 @@ class FeatureView(TableDescriptor, ABC):
             raise FeathubException(
                 "Build this feature view before getting output fields."
             )
-        output_fields = []
+
+        feature_fields = [feature.name for feature in self.get_resolved_features()]
         if self.keep_source_fields:
-            output_fields.extend(source_fields)
-        elif self.timestamp_field is not None:
-            output_fields.append(self.timestamp_field)
-
-        for feature in self.get_resolved_features():
-            if feature.keys is not None:
-                output_fields.extend(feature.keys)
-            output_fields.append(feature.name)
-
-        # Order output fields similar to their order in the source table.
-        reordered_output_fields = []
-        for field in source_fields:
-            if field in output_fields:
-                reordered_output_fields.append(field)
-        reordered_output_fields.extend(output_fields)
-
-        return list(OrderedDict.fromkeys(reordered_output_fields))
+            extra_fields = list(set(source_fields) - set(feature_fields))
+        else:
+            timestamp_and_keys = (
+                [self.timestamp_field] if self.timestamp_field is not None else []
+            )
+            for feature in self.get_resolved_features():
+                if feature.keys is not None:
+                    timestamp_and_keys.extend(feature.keys)
+            extra_fields = list(set(timestamp_and_keys) - set(feature_fields))
+        # Order extra fields in the same order as their position in source.
+        extra_fields = [f for f in source_fields if f in extra_fields]
+        return extra_fields + feature_fields
 
     def get_output_features(self) -> List[Feature]:
         if self.is_unresolved():
--- a/python/feathub/feature_views/sliding_feature_view.py
+++ b/python/feathub/feature_views/sliding_feature_view.py
@@ -11,7 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import Dict, Union, Sequence, Set, Any, Optional
+from typing import Dict, Union, Sequence, Set, Any, Optional, List, OrderedDict
 
 from feathub.common import types
 from feathub.common.config import BaseConfig, ConfigDef
@@ -192,6 +192,24 @@ class SlidingFeatureView(FeatureView):
             transform=WINDOW_TIME_EXPR,
         )
         return window_time_feature
+
+    def get_output_fields(self, source_fields: List[str]) -> List[str]:
+        if self.is_unresolved():
+            raise FeathubException(
+                "Build this feature view before getting output fields."
+            )
+
+        key_fields = [
+            key
+            for feature in self.get_resolved_features()
+            if feature.keys is not None
+            for key in feature.keys
+        ]
+        output_fields = [f for f in source_fields if f in key_fields]
+        output_fields.append(self.timestamp_field)
+        output_fields.extend([feature.name for feature in self.get_resolved_features()])
+
+        return list(OrderedDict.fromkeys(output_fields))
 
     def build(
         self, registry: Registry, props: Optional[Dict] = None
--- a/python/feathub/feature_views/tests/test_feature_view.py
+++ b/python/feathub/feature_views/tests/test_feature_view.py
@@ -15,7 +15,7 @@
 import unittest
 
 from feathub.common.exceptions import FeathubException
-from feathub.common.types import Int64
+from feathub.common.types import Int64, String
 from feathub.feature_tables.sources.datagen_source import DataGenSource
 from feathub.feature_tables.sources.file_system_source import FileSystemSource
 from feathub.feature_views.feature import Feature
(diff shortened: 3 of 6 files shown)


python/feathub/feature_views/feature_view.py
```diff
@@ -133,25 +133,28 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
...
             )
-        output_fields = []
+
+        derived_fields = [feature.name for feature in self.get_resolved_features()]
+        key_fields = [
+            key
+            for feature in self.get_resolved_features()
+            if feature.keys
+            for key in feature.keys
+        ]
```
yunfengzhou-hub, line 143: If there are multiple features, `key_fields` might contain duplicate entries.

Besides, given that derived feature view can have features with different keys, it might be unnecessary to consider key fields when ordering fields.

python/feathub/feature_views/feature_view.py
```diff
@@ -133,25 +133,28 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
             raise FeathubException(
                 "Build this feature view before getting output fields."
             )
-        output_fields = []
+
+        derived_fields = [feature.name for feature in self.get_resolved_features()]
```
yunfengzhou-hub, line 137: How about moving this part of logic to `DerivedFeatureView`? `derived_fields` only exists in `DerivedFeatureView` and different views can have different ordering strategies.

python/feathub/feature_views/tests/test_derived_feature_view.py
```diff
@@ -52,6 +52,7 @@ def test_derived_feature_view_filter_expr(self):
         expected_result_df = expected_result_df.sort_values(by=["time"]).reset_index(
             drop=True
         )
+        expected_result_df = expected_result_df[result_df.columns]
```
yunfengzhou-hub, line 55: We may not need to reorder the fields of `expected_result_df` according to the previous comments. And if we have to reorder the fields, it might be better to modify the initialization of `expected_result_df` directly.

yunfengzhou-hub (commented):

python/feathub/feature_views/tests/test_feature_view.py
```diff
@@ -191,3 +192,47 @@ def test_duplicate_feature_names_throw_exception(self):
...
+
+        built_feature_view_1 = feature_view_1.build(self.registry)
+
+        if isinstance(built_feature_view_1, FeatureView):
+            self.assertEqual(
+                ["id", "val2", "val3", "val1"],
+                built_feature_view_1.get_output_fields(field_names),
+            )
+        else:
+            raise Exception("FeatureView is expected.")
```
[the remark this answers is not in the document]
jiangxin369: The Exception is to ensure that the test would fail if `built_feature_view_1` is not as expected. Adding the `if instanceOf...else...` is to avoid mypy check.

python/feathub/feature_views/feature_view.py
jiangxin369: The following `OrderedDict.fromkeys()` would handle the duplicate entries.

python/feathub/feature_views/feature_view.py
jiangxin369: It is called derived_fields but not only used in `DerivedFeatureView`, I renamed it with `feature_fields`.

python/feathub/feature_views/tests/test_feature_view.py
```diff
@@ -191,3 +192,47 @@ def test_duplicate_feature_names_throw_exception(self):
                 keep_source_fields=True,
             )
         self.assertIn("contains duplicated feature name", ctx.exception.args[0])
+
+    def test_get_output_fields(self):
+        field_names = ["id", "val1", "val2", "val3"]
+        source = DataGenSource(
+            name="source_1",
+            schema=Schema(field_names, [Int64, Int64, Int64, Int64]),
+            timestamp_field="lpep_dropoff_datetime",
```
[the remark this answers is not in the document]
jiangxin369: Yes, I created https://github.com/alibaba/feathub/issues/103.

python/feathub/feature_views/tests/test_derived_feature_view.py
jiangxin369: The `expected_result_df` is usually a copy of the input data, we don't initialize it. And we already have tests to ensure the order of feature view is correct, so we don't have to construct the `expected_result_df` with field names explicitly. If you're worrying about the risk of the different numbers of columns, I can check the number of columns before adjusting the order of columns. How do you think？

python/feathub/feature_views/feature_view.py
```diff
@@ -133,25 +133,35 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
...
+                key
+                for feature in self.get_resolved_features()
+                if feature.keys is not None
+                for key in feature.keys
+            ]
+            reserved_fields.extend(key_fields)
+            reserved_fields_not_in_features = [
+                f
+                for f in source_fields
+                if f in reserved_fields and f not in feature_fields
```
yunfengzhou-hub, line 159: It is supposed that all timestamp or key fields are source fields. Thus how about writing it this way?
```python
[f for f in reserved_fields if f not in feature_fields]
```

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -193,6 +193,24 @@ def _get_window_time_feature(
...
+                "Build this feature view before getting output fields."
+            )
+
+        key_fields = [
+            key
+            for feature in self.get_resolved_features()
+            if feature.keys is not None
+            for key in feature.keys
+        ]
+        output_fields = [f for f in source_fields if f in key_fields]
```
yunfengzhou-hub, line 208: Same as above.

python/feathub/feature_views/tests/test_feature_view.py
```diff
@@ -201,3 +202,80 @@ def test_duplicate_feature_names_throw_exception(self):
...
+        )
+
+        feature_view_1 = DerivedFeatureView(
+            name="feature_view_1",
+            source=source,
+            features=["time", feature_1, feature_2],
+            keep_source_fields=True,
+        )
+
+        built_feature_view_1 = feature_view_1.build(self.registry)
```
yunfengzhou-hub, line 231: In `test_kafka_source_sink.py` there is a following code snippet:
```python
            flink_table_descriptor: NativeFlinkTableDescriptor = (
                from_descriptor.call_args[0][0]
            )
```

Similarly, we may also write the code here this way
```python
built_feature_view_1: FeatureView = feature_view_1.build(self.registry)
```

and remove the `raise Exception("FeatureView is expected.")` below.

yunfengzhou-hub (commented):
Thanks for the update. Left some minor comments as below.

python/feathub/feature_views/feature_view.py
jiangxin369: Traversing in `source_fields` is to make the order of timestamp or key fields the same as those in `source_fields`.

python/feathub/feature_views/tests/test_feature_view.py
jiangxin369: Note that the `if... else raise Exception` code block is to avoid `mypy` type check failure, seems the above code snippet doesn't solve the problem. I found that `mypy` can skip checking some lines if users make sure that the lines are correct by adding a note like `# type: ignore`. Do you think it is enough for a test?

yunfengzhou-hub: Thanks for the update. This solution is also acceptable to me.

yunfengzhou-hub:

python/feathub/feature_views/feature_view.py
```diff
@@ -133,25 +133,35 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
             raise FeathubException(
                 "Build this feature view before getting output fields."
             )
-        output_fields = []
+
+        feature_fields = [feature.name for feature in self.get_resolved_features()]
```
lindong28, line 137: Can you update the Python doc to explicitly describe the order of fields emitted by the FeatureView?

And should we put features explicitly specified by users before features that are implicitly included (e.g. keys, timestsamp)? This approach allows users to easily determine the positions of features that are explicitly specified.

lindong28 (commented):
Thanks for the PR. Left one comment below.

python/feathub/feature_views/feature_view.py
jiangxin369: I updated the Python doc.

I thought about putting features before reserved fields, but users are more accustomed to having the key fields before others. And users do not access fields by positions, right? If users really need a specific position, they can add a DerivedFeatureView to select all fields in their order.

lindong28: Position of fields matters here because otherwise we won't need this PR.

The concern with putting keys before user-specified features is that users won't have full control of the order of fields emitted by the FeatureView.

Using an extra DerivedFeatureView is more complicated than using the existing FeatureView to get the expected orders.

jiangxin369: According to the offline discussion, we agree that the order of keys and features doesn't matter currently. But the docs need to be clear.

I've updated the Python doc again, please have a check.

python/feathub/feature_views/feature_view.py
```diff
@@ -133,25 +137,35 @@ def get_output_fields(self, source_fields: List[str]) -> List[str]:
...
```
lindong28, line 141: Would the following code be shorter and a bit more readable?

```
feature_fields = [feature.name for feature in self.get_resolved_features()]
if self.keep_source_fields:
    extra_fields = list(set(source_fields) - set(feature_fields))
else:
    timestamp_and_keys = []
    for feature in self.get_resolved_features():
        if feature.keys is not None:
            timestamp_and_keys.extend(feature.keys)
    if self.timestamp_field is not None:
        timestamp_and_keys.append(self.timestamp_field)
    extra_fields = list(set(timestamp_and_keys) - set(feature_fields))
# Order extra fields in the same order as their position in source.
extra_fields = [
    f for f in source_fields if f in extra_fields
]
return extra_fields + feature_fields
```

jiangxin369: Yes, it's simpler. But the `set` doesn't keep order, but we want the `extra_fields` in the order as they appear in the source fields.

lindong28: Would the following code snippet ensure that the extra_fields have the same order as `source_fields`?

```
# Order extra fields in the same order as their position in source.
extra_fields = [
    f for f in source_fields if f in extra_fields
]
```

jiangxin369: Yes, you're right, I lost it.

lindong28 (commented):
Thanks for the update. LGTM.

PR #107, opened 2023-03-29
Support CASE and filtering in SparkProcessor

yunfengzhou-hub:

This pull request adds support for expressions like `CASE` and filtering in SparkProcessor's ExpressionTransform and DerivedFeatureView.

## Brief change log

- Supports parsing `CASE`, `IS`/`IS NOT`, and `NULL` in SparkProcessor's expression.
- Supports filtering SparkProcessor's DerivedFeatureView with expressions.
- Supports filtering SparkProcessor's DerivedFeatureView with start/end time.

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/processors/spark/dataframe_builder/spark_dataframe_builder.py
+++ b/python/feathub/processors/spark/dataframe_builder/spark_dataframe_builder.py
@@ -11,7 +11,8 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import Dict, Tuple, List, Sequence, Union
+from datetime import datetime
+from typing import Dict, Tuple, List, Sequence, Union, Optional
 
 import pandas as pd
 
@@ -78,6 +79,8 @@ class SparkDataFrameBuilder:
         self,
         features: TableDescriptor,
         keys: Union[pd.DataFrame, TableDescriptor, None] = None,
+        start_datetime: Optional[datetime] = None,
+        end_datetime: Optional[datetime] = None,
     ) -> NativeSparkDataFrame:
         """
         Convert the given features to native Spark DataFrame.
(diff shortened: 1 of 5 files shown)


python/feathub/processors/spark/dataframe_builder/time_utils.py
```diff
@@ -46,7 +46,7 @@ def append_unix_time_attribute_column(
         return df.withColumn(
             EVENT_TIME_ATTRIBUTE_NAME,
             functions.expr(
-                f"unix_millis(to_timestamp("
-                f"`{timestamp_field}`,'{java_datetime_format}'))"
+                f"unix_millis(from_utc_timestamp(to_timestamp("
```
Sxnan, line 49: Can you check if this method can handle the case where the input contains information about the timezone? If not, we may need to add a TODO for that case, similar to python/feathub/processors/flink/table_builder/source_sink_utils_common.py#64

Sxnan (commented):
Thanks for the PR. Just left one comment.

Sxnan:
@yunfengzhou-hub Thanks for the update! LGTM. @lindong28 Could you take a look at this PR?

python/feathub/processors/spark/dataframe_builder/spark_dataframe_builder.py
```diff
@@ -322,11 +340,48 @@ def _get_dataframe_from_derived_feature_view(
                     f"{type(feature.transform).__name__} for feature {feature.name}."
                 )
 
+        tmp_dataframe = self._apply_filter_if_any(
```
lindong28, line 343: Since this method is only meaningful when `filter_expr` is not None, it seems more reasonable to invoke it only when `filter_expr` is not None.

Also, since this method is only invoked in one place and it is simpler, maybe we should just inline it.

python/feathub/processors/spark/dataframe_builder/spark_dataframe_builder.py
```diff
@@ -322,11 +340,48 @@ def _get_dataframe_from_derived_feature_view(
...
+    def _apply_filter_if_any(
+        dataframe: NativeSparkDataFrame, filter_expr: Optional[str]
+    ) -> NativeSparkDataFrame:
+        # TODO: Apply filtering as early as possible to improve performance.
+        if filter_expr is None:
+            return dataframe
+        return dataframe.filter(functions.expr(to_spark_sql_expr(filter_expr)))
+
+    @staticmethod
+    def _range_dataframe_by_time(
```
lindong28, line 362: Would it be more intuitive to name it `_filter_dataframe_by_time`?

lindong28 (commented):
Thanks for the PR. LGTM overall. Left some minor comments.

yunfengzhou-hub:

lindong28:
Thanks for the PR. LGTM.

PR #108, opened 2023-03-29
Fix FlinkProcessor SlidingWindow deduplication logic with limit

yunfengzhou-hub:

This pull request fixes the bug that FlinkProcessor cannot correctly deduplicate sliding window aggregation results when `limit` is not null.

## Brief change log

  - fixes the bug that FlinkProcessor cannot correctly deduplicate sliding window aggregation result when `limit` is not null.

  - The public API: no

## Documentation


Sxnan (commented):
Thanks for the PR! LGTM. @lindong28 Could you take another look?

lindong28 (commented):
Thanks for the PR! LGTM.

PR #110, opened 2023-03-30
LocalProcessor supports filtering in Derived/SlidingFeatureView

Sxnan:

Update LocalProcessor to support filtering in DerivedFeatureView and SlidingFeatureView.

## Brief change log

- Update LocalProcessor to support filtering in DerivedFeatureView and SlidingFeatureView
- Fix the bug in the LocalProcessor ast evaluator to handle IS NULL for numpy floating and python native float type

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
+++ b/python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
@@ -155,7 +155,17 @@ class LocalAstEvaluator(AbstractAstEvaluator):
 
     def eval_is_op(self, ast: IsOp, variables: Optional[Dict]) -> Any:
         left_value = self.eval(ast.left_child, variables)
-        is_none = left_value is None or np.isnan(left_value)
+        is_none = left_value is None
+
+        # Pandas converts None value of numeric type to NAN. Therefore, it is unclear
+        # whether NAN represents None or NAN. We treat all the NAN as None here as it
+        # should be the common case. This may result in a different behavior from
+        # Spark/Flink processor where they have NULL to represent None value.
+        # TODO: Add docs for LocalProcessor and warn about the different behavior from
+        #  Spark/Flink processor.
+        if isinstance(left_value, float) or isinstance(left_value, np.generic):
+            is_none = np.isnan(left_value)
+
         if ast.is_not:
             return not is_none
 
--- a/python/feathub/processors/local/ast_evaluator/tests/test_local_ast_evaluator.py
+++ b/python/feathub/processors/local/ast_evaluator/tests/test_local_ast_evaluator.py
@@ -15,6 +15,8 @@
 import unittest
 from datetime import datetime
 
+import numpy as np
+
 from feathub.common.exceptions import FeathubException, FeathubExpressionException
 from feathub.dsl.expr_parser import ExprParser
 from feathub.processors.local.ast_evaluator.local_ast_evaluator import LocalAstEvaluator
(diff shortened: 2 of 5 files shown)


Sxnan:

python/feathub/processors/local/ast_evaluator/tests/test_local_ast_evaluator.py
```diff
@@ -112,6 +114,9 @@ def test_is_op(self):
         self.assertEqual(False, self._eval("a IS NULL", {"a": 1}))
         self.assertEqual(False, self._eval("a IS NOT NULL", {"a": None}))
         self.assertEqual(True, self._eval("a IS NOT NULL", {"a": 1}))
+        self.assertEqual(False, self._eval("a IS NOT NULL", {"a": np.NAN}))
```
yunfengzhou-hub, line 117: Could you please also add an e2e test in `DerivedFeatureViewITTest` that evaluates `NAN` is/is not `NULL`? I hope to be guaranteed that judging `NAN` as `NULL` is widely accepted in Flink and Spark as well.

python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
```diff
@@ -155,7 +155,10 @@ def eval_group_node(self, ast: GroupNode, variables: Optional[Dict]) -> Any:
 
     def eval_is_op(self, ast: IsOp, variables: Optional[Dict]) -> Any:
         left_value = self.eval(ast.left_child, variables)
-        is_none = left_value is None or np.isnan(left_value)
+        is_none = left_value is None
+        if isinstance(left_value, float) or isinstance(left_value, np.floating):
```
yunfengzhou-hub, line 159: Do we need to judge from `np.int8` to `np.float64`?

python/feathub/processors/local/local_processor.py
```diff
@@ -611,3 +612,7 @@ def _get_dependent_features(self, feature_view: FeatureView) -> Sequence[Feature
             if feature not in dependent_features:
                 dependent_features.append(feature)
         return dependent_features
+
+    def _filter_df(self, df: pd.DataFrame, filter_expr: str) -> pd.DataFrame:
```
yunfengzhou-hub, line 616: Would it be better to rename this method to `_filter_dataframe`? I noticed that none of the other methods use abbreviations in its signature.

yunfengzhou-hub (commented):

python/feathub/processors/local/ast_evaluator/local_ast_evaluator.py
Sxnan: Thanks for the suggestion. We can check if the `left_value` is an instance of `np.generic`. This takes care of all the numpy dtypes.

python/feathub/processors/local/ast_evaluator/tests/test_local_ast_evaluator.py
Sxnan: I think Flink and Spark don't have the notion of NAN. I added the test case where a float-type column has a null value.

Sxnan:
@yunfengzhou-hub Thanks for your comments! I updated the PR. Could you take another look?

python/feathub/processors/local/ast_evaluator/tests/test_local_ast_evaluator.py
yunfengzhou-hub: I noticed that Spark SQL has the notion of NAN[1], which is a numeric value and not equal to null. It might be better to follow this practice in LocalProcessor. Maybe we can add some function named `ISNAN()` to judge the corresponding situation in future.

[1] https://spark.apache.org/docs/3.3.1/sql-ref-datatypes.html#nan-semantics

yunfengzhou-hub:
Thanks for the update. LGTM. @lindong28 could you please take another look?

Sxnan:
@lindong28 I updated the PR according to our offline discussion. Could you please take another look?

lindong28:
Thanks for the PR. LGTM.

PR #111, opened 2023-03-30
Raise exception if the timestamp_field is specified but not found in the schema

jiangxin369:

Check if the schema contains timestamp field and key fields if they are set.

## Brief change log

  - Check if the schema contains timestamp field and key fields if they are set.

  - The public API: (no)

## Documentation

### DIFF
--- a/python/feathub/feature_tables/feature_table.py
+++ b/python/feathub/feature_tables/feature_table.py
@@ -74,6 +74,23 @@ class FeatureTable(TableDescriptor, ABC):
         self.properties = properties
         self.schema = schema
 
+        if schema is not None:
+            if (
+                timestamp_field is not None
+                and timestamp_field not in schema.field_names
+            ):
+                raise FeathubException(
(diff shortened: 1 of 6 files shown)


jiangxin369:
@Sxnan Could you help review this PR?

python/feathub/feature_tables/feature_table.py
```diff
@@ -74,6 +74,20 @@ def __init__(
         self.properties = properties
         self.schema = schema
 
+        if schema:
```
Sxnan, line 77: Is it safer if we explicitly check `if schema is not None`? Same for the other None check.

Sxnan (commented):
Thanks for the PR! Just left one minor comment below.

Sxnan:
Thanks for your update! LGTM. @lindong28 Could you help take a look at this PR?

lindong28:
Thanks for the PR. LGTM.

PR #116, opened 2023-04-03
Update validations of Spark/FlinkProcessor and FeatureView 

### LINKED ISSUE #119: Apply Flink native configuration before creating tableEnv instance
In an attempt to create FeatHubClient with the following code:
```python
client = FeathubClient(
    props={
        "processor": {
            "type": "flink",
            "flink": {
                "deployment_mode": "cli",
                "native.execution.runtime-mode": "BATCH"
            },
        },
        "registry": {
            "type": "local",
            "local": {
                "namespace": "default",
            },
        },
        "feature_service": {
            "type": "local",
            "local": {},
        },
    }
)
```

The following error is thrown when the program arrives at `Table#to_pandas`:
```
java.lang.IllegalArgumentException: Mismatch between configured runtime mode and actual runtime mode. Currently, the 'execution.runtime-mode' can only be set when instantiating the table environment. Subsequent changes are not supported. Please instantiate a new TableEnvironment if necessary.
```

Thus the configuration initialization process should be improved to avoid exceptions like this.

Sxnan:

This PR updates the validation logic of Spark/FlinkProcessor and FeatureView.

## Brief change log

- Update Flink/SparkProcessor to report error when processor config conflict with native.* config
- Introduce class to keep track of built-in functions
- Validate features order in Sliding/DerivedFeatureView
- Fix #119

  - The public API: no

## Documentation


python/feathub/processors/spark/spark_processor.py
```diff
@@ -35,6 +35,7 @@
     SparkProcessorConfig,
```
yunfengzhou-hub, line 35: Hi @Sxnan, could you please fix the todo in SparkProcessorConfig file and add a validator for `MASTER_CONFIG`? This TODO had been assigned to me according to offline discussions, but I think it is also proper for this PR to cover this TODO.

python/feathub/dsl/ast.py
```diff
@@ -234,12 +225,8 @@ def __init__(self, func_name: str, args: ArgListNode) -> None:
 
     def eval_dtype(self, variable_types: Dict[str, DType]) -> DType:
         arg_types = [arg.eval_dtype(variable_types) for arg in self.args.values]
-        function_type_evaluator = FUNCTION_RESULT_TYPE_EVALUATOR.get(self.func_name)
-        if function_type_evaluator is None:
-            raise RuntimeError(
-                f"Unknown result type evaluator for function {self.func_name}."
-            )
-        return function_type_evaluator(arg_types)
+        builtin_func = BuiltInFunc(self.func_name)
```
lindong28, line 228: Instead of constructing an instance of `BuiltInFunc` by calling its constructor, would it be more intuitive to use a static method such as `BuiltInFunc#get(func_name)`?

lindong28 (commented):

python/feathub/dsl/expr_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+    lexer.input(feathub_expr)
+    variables = []
+    while True:
+        tok = lexer.token()
+        if not tok:
+            break
+        if tok.type == "ID" and tok.value not in BUILT_IN_FUNC_NAMES:
+            variables.append(tok.value)
+    return set(variables)
```
yunfengzhou-hub, line 36: how about creating `variables` as a `set` variable from the beginning?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -107,6 +108,42 @@ def build(
...
 
+    @staticmethod
+    def _validate(features: List[Feature], source: TableDescriptor) -> None:
+        # Check if the feature only depends on the features specified earlier or the
+        # features in the source table.
+        valid_variables = set([f.name for f in source.get_output_features()])
+        for feature in features:
+            transform = feature.transform
+            if isinstance(transform, JoinTransform):
+                variables = set()
```
yunfengzhou-hub, line 119: Is it possible to get the name of the table to be joined and consider its fields as well?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,46 +277,88 @@ def _validate(features: Sequence[Union[str, Feature]]) -> None:
 
         # Per-row transform features listed before the first SlidingWindowTransform
         # feature.
-        per_row_transform_feature_names: Set[str] = set()
-        sliding_window_transforms_group_by_keys: Set[str] = set()
+        pre_sliding_per_row_feature_names: Set[str] = set()
+
+        # Per-row transform features listed after the first SlidingWindowTransform
+        # feature.
+        post_sliding_per_row_features: Set[Feature] = set()
```
yunfengzhou-hub, line 284: Would it be more readable to use either `Set[str]` or `Set[Feature]` for both sets above?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,46 +277,88 @@ def _validate(features: Sequence[Union[str, Feature]]) -> None:
...
+        sliding_features_group_by_key: Set[str] = set()
+        sliding_features_variables: Set[str] = set()
 
         for feature in features:
             if isinstance(feature, str):
-                if len(sliding_window_transforms_group_by_keys) == 0:
-                    per_row_transform_feature_names.add(feature)
+                if len(sliding_features_group_by_key) == 0:
+                    pre_sliding_per_row_feature_names.add(feature)
                 continue
```
yunfengzhou-hub, line 294: Should we try to handle string transformations after sliding window transform?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -254,8 +262,7 @@ def to_json(self) -> Dict:
             "filter_expr": self.filter_expr,
         }
 
-    @staticmethod
-    def _validate(features: Sequence[Union[str, Feature]]) -> None:
+    def _validate(self, features: Sequence[Union[str, Feature]]) -> None:
```
yunfengzhou-hub, line 265: Should we check that pre-sliding features and sliding features only use fields from the source table, like what this PR validates in DerivedFeatureView?

python/feathub/dsl/built_in_func.py
```diff
@@ -0,0 +1,41 @@
...
+    def get(func_name: str) -> "BuiltInFunc":
+        return BuiltInFunc(func_name)
+
+    def get_result_type(self, input_types: List[DType]) -> DType:
+        """
+        Get the result data type of the builtin function.
+        """
+        if self == BuiltInFunc.LOWER:
+            return String
+        elif self == BuiltInFunc.UNIX_TIMESTAMP:
```
yunfengzhou-hub, line 35: I'm not sure enum would be a good design for built-in functions, since we may have hundreds of functions to support. Would it be better to adopt another design pattern, like `org.apache.flink.table.functions.BuiltInFunctionDefinitions`?

yunfengzhou-hub (commented):

python/feathub/feature_views/derived_feature_view.py
Sxnan: The `variable` here refers to the input feature to compute the new feature. We then check that the `variable`  is a subset of `valid_variables`.  For JoinTransform, it doesn't has an input feature to compute the new feature. We only need to add the join feature to the `valid_variables`.

python/feathub/feature_views/sliding_feature_view.py
Sxnan: String transformations can only refer to a feature from the source table. Thus, there should be no string transformation after the sliding window transform.

yunfengzhou-hub: I agree with it that there should be no string transformation after the sliding window transform. And if such cases appeared, FeatHub should throw exceptions instead of ignoring it.

Sxnan: The validation is now done after the string feature is resolved. It should not be a problem now.

python/feathub/feature_views/tests/test_sliding_feature_view.py
```diff
@@ -47,8 +47,8 @@ def setUp(self):
             path="dummy_source_file",
             data_format="csv",
             schema=Schema(
-                ["id", "fare_amount", "lpep_dropoff_datetime"],
-                [types.Int32, types.Int32, types.Int64],
+                ["id", "id2", "fare_amount", "distance", "lpep_dropoff_datetime"],
```
yunfengzhou-hub, line 50: It seems that "id2" is not used in the test cases. Would it be better to remove it?

yunfengzhou-hub (commented):
Thanks for the update. Left some comments as below.

python/feathub/feature_views/tests/test_sliding_feature_view.py
Sxnan: It is used in line 255.

yunfengzhou-hub:

lindong28 (commented):
Thanks for the PR. LGTM overall. Left some minor comments.

Sxnan:

lindong28:
Thanks for the update! LGTM.

PR #117, opened 2023-04-06
Add feature embedding and fraud detection notebook example

yunfengzhou-hub:

Add feature embedding and fraud detection notebook example

## Brief change log

- Adds feature-embedding example
- Adds fraud detection example
- Adds instructions to configure python environment in notebook examples.
- Uses FlinkProcessor in nyc taxi example.
- Makes a few bugfixes and improvements to make sure the examples can be executed.
  - Introduces master config in FlinkProcessor.
  - Adds support for count agg func in OverWindowTransform in FlinkProcessor.
  - Fixes ExpressionTransform's bug when feature name equals reserved keyword.
  - Fixes flink_table_builder's bug that ignores timestamp field configuration.
  - Fixes bug that feature view's key cannot be correctly derived when source's key is none and feature has keys.

  - The public API: no

## Documentation

### DIFF
--- a/docs/examples/feature_embedding.ipynb
+++ b/docs/examples/feature_embedding.ipynb
@@ -1,0 +1,455 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[![Open In\n",
+    "Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/alibaba/feathub/blob/master/docs/examples/feature_embedding.ipynb)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Feature Embedding\n",
+    "\n",
+    "Feature embedding is a way to translate a high-dimensional feature vector to a\n",
+    "lower-dimensional vector, where the embedding can be learned and reused across\n",
+    "models. In this example, we show how one can define feature embeddings via\n",
+    "Python UDF (User Defined Function).\n",
+    "\n",
+    "We use a sample hotel review dataset downloaded from [Azure-Samples\n",
+    "repository](https://github.com/Azure-Samples/azure-search-sample-data). The\n",
+    "original dataset can be found\n",
+    "[here](https://www.kaggle.com/datasets/datafiniti/hotel-reviews).\n",
+    "\n",
+    "For the embedding, a pre-trained [HuggingFace Transformer\n",
+    "model](https://huggingface.co/sentence-transformers) is used to encode texts\n",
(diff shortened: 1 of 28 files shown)


docs/examples/feature_embedding.ipynb
```diff
@@ -0,0 +1,432 @@
...
+    "but building such models is out of scope and thus we don't cover that in this\n",
+    "notebook."
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Install dependencies\n",
```
lindong28, line 42: Would it be better to uncomment these code so that users can install the dependencies easily by just clicking the cell?

docs/examples/feature_embedding.ipynb
```diff
@@ -0,0 +1,432 @@
+{
+ "cells": [
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "[![Open In\n",
+    "Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/alibaba/feathub/blob/master/docs/examples/feature_embedding.ipynb)"
```
lindong28, line 9: Can you help make the capability of running the example in CoLab more obvious, e.g. by adding text describing this capability?

docs/examples/feature_embedding.ipynb
```diff
@@ -0,0 +1,432 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "client = FeathubClient(\n",
+    "    props={\n",
+    "        \"processor\": {\n",
+    "            \"type\": \"flink\",\n",
+    "            \"flink\": {\n",
+    "                \"deployment_mode\": \"cli\",\n",
```
lindong28, line 156: Should we use `session mode` with IP = local?

And can we combine `rest.address` and `rest.port` into one config named `master` similar to Spark?

python/dev-requirements.txt
```diff
@@ -27,3 +27,6 @@ types-redis~=4.3.21
 types-tzlocal~=4.2
 cryptography
 jupyter
+sentence-transformers
```
lindong28, line 30: Do we still need to specify these dependencies in dev-requirement if the Notebook examples already have the code to install these dependencies?

lindong28 (commented):

docs/examples/feature_embedding.ipynb
yunfengzhou-hub: `rest.address` and `rest.port` is more friendly to users that have been familiar with Flink, so it might be better to keep them instead of introducing `master` configuration. I have add supports that when `rest.address` is `"local"`, FlinkProcessor would submit the job to a mini cluster.

lindong28:
Thanks for the update. LGTM.

PR #125, opened 2023-04-23
Support existing connectors with Spark local mode in LocalProcessor

yunfengzhou-hub:

This pull request supports in LocalProcessor all the connectors that have been supported in SparkProcessor.

## Brief change log

- Add support to read/write with SparkProcessor-supported connectors in LocalProcessor with Spark local mode.
- Add test for `DataGenSource`, `PrintSink` and `BlackHoleSink` in LocalProcessorITTest.

## Verifying the changes

- Most changes in this PR has been covered by the added it tests.
- LocalProcessor's ability to read/write HDFS is covered in the e2e example [here](https://github.com/yunfengzhou-hub/feathub-examples/tree/local-hdfs-example/local-read-write-hdfs).

  - The public API: no

## Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,11 @@ See [Basic Concepts](docs/content/concepts/basic-concepts.md) for more details a
 FeatHub supports the following compute engines to execute feature ETL pipeline:
 - [Apache Flink 1.15](docs/content/engines/flink.md)
 - [Aapche Spark 3.3](docs/content/engines/spark.md)
-- Local processor, which is implemented using the Pandas library and computes features in the given Python process.
+- Local processor, which is implemented using the Pandas library and computes
+  features in the given Python process. Additionally, if the
+  feathub-nightly[spark] is installed, the Local processor can utilize Spark's
+  local mode for accessing storages (e.g. HDFS) that it otherwise did not
+  support.
 
 
 ## FeatHub SDK Highlights
@@ -213,7 +217,8 @@ $ python -m pip install --upgrade feathub-nightly
 # Run the following command if you plan to use Apache Flink cluster
 $ python -m pip install --upgrade "feathub-nightly[flink]"
 
-# Run the following command if you plan to use Apache Spark cluster
+# Run the following command if you plan to use Apache Spark cluster, or to use
+# Spark-supported storage in a local process. 
 $ python -m pip install --upgrade "feathub-nightly[spark]"
 ```
 
--- a/python/feathub/processors/local/local_processor.py
+++ b/python/feathub/processors/local/local_processor.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timedelta
 from typing import (
     Dict,
@@ -20,6 +21,7 @@ from typing import (
     Any,
     Sequence,
 )
+from urllib.parse import urlparse
 
 import pandas as pd
 from dateutil.tz import tz
@@ -30,8 +32,11 @@ from feathub.common.exceptions import FeathubException, FeathubTransformationExc
 from feathub.common.types import to_numpy_dtype
 from feathub.dsl.expr_parser import ExprParser
 from feathub.feature_tables.feature_table import FeatureTable
+from feathub.feature_tables.sinks.black_hole_sink import BlackHoleSink
 from feathub.feature_tables.sinks.file_system_sink import FileSystemSink
 from feathub.feature_tables.sinks.memory_store_sink import MemoryStoreSink
+from feathub.feature_tables.sinks.print_sink import PrintSink
+from feathub.feature_tables.sources.datagen_source import DataGenSource
 from feathub.feature_tables.sources.file_system_source import FileSystemSource
 from feathub.feature_views.derived_feature_view import DerivedFeatureView
 from feathub.feature_views.feature import Feature
@@ -77,6 +82,27 @@ from feathub.table.schema import Schema
 from feathub.table.table_descriptor import TableDescriptor
 
 
+def _is_spark_supported_source(source: FeatureTable) -> bool:
+    return isinstance(source, (FileSystemSource, DataGenSource))
+
+
+def _is_spark_supported_sink(sink: FeatureTable) -> bool:
+    return isinstance(
+        sink,
+        (
+            FileSystemSink,
+            PrintSink,
+            BlackHoleSink,
+            MemoryStoreSink,
+        ),
+    )
+
+
+def _is_local_file_or_dir(url: str) -> bool:
+    url_parsed = urlparse(url)
+    return url_parsed.scheme in ("file", "")
+
+
 class LocalProcessor(Processor):
     """
     A LocalProcessor uses CPUs on the local machine to compute features and uses Pandas
@@ -92,11 +118,14 @@ class LocalProcessor(Processor):
         self.props = props
         self.registry = registry
 
-        config = LocalProcessorConfig(props)
-        self.timezone = tz.gettz(config.get(TIMEZONE_CONFIG))
+        self.config = LocalProcessorConfig(props)
+        self.timezone = tz.gettz(self.config.get(TIMEZONE_CONFIG))
 
         self.parser = ExprParser()
         self.ast_evaluator = LocalAstEvaluator(tz=self.timezone)
+
+        self.spark_session: Optional[Any] = None
+        self.executor = ThreadPoolExecutor()
 
     def get_table(
         self,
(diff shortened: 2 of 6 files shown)


python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +211,41 @@ def materialize_features(
                 timestamp_field=features.timestamp_field,
                 timestamp_format=features.timestamp_format,
             )
-        elif isinstance(sink, FileSystemSink):
+        elif isinstance(sink, FileSystemSink) and _is_local_file_or_dir(sink.path):
```
Sxnan, line 214: Let's update the docs of the FileSystem connector as the local processor supports HDFS now.

Sxnan (commented):
Thanks for the PR! Just left one comment below.

Sxnan:
LGTM! @lindong28 Could you take another look at this PR?

python/feathub/processors/local/local_processor.py
```diff
@@ -20,9 +21,11 @@
     Any,
     Sequence,
 )
+from urllib.parse import urlparse
 
 import pandas as pd
 from dateutil.tz import tz
+from pyspark.sql import SparkSession
```
lindong28, line 28: Can we import Spark-related dependencies dynamically only if we need to use them?

Since we do not require users to always install feathub[spark] to use local processor, it will be useful to catch and report proper error (e.g. please install FeatHub with Spark to use xxx) if pyspark is not installed.

README.md
```diff
@@ -121,7 +121,8 @@ See [Basic Concepts](docs/content/concepts/basic-concepts.md) for more details a
 FeatHub supports the following compute engines to execute feature ETL pipeline:
 - [Apache Flink 1.15](docs/content/engines/flink.md)
 - [Aapche Spark 3.3](docs/content/engines/spark.md)
-- Local processor, which is implemented using the Pandas library and computes features in the given Python process.
+- Local processor, which is implemented using the Pandas library and Spark local
```
lindong28, line 124: Let's not forcing users to always use pyspark to use local processor. How about this:

Local processor, which is implemented using the Pandas library and computes features in the given Python process. Additionally, if the feathub-nightly[spark] is installed, the Local processor can utilize pyspark local mode for accessing storages (e.g. HDFS) that it otherwise did not support.

python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +211,41 @@ def materialize_features(
                 timestamp_field=features.timestamp_field,
                 timestamp_format=features.timestamp_format,
             )
-        elif isinstance(sink, FileSystemSink):
+        elif isinstance(sink, FileSystemSink) and _is_local_file_or_dir(sink.path):
             insert_into_file_sink(features_df, sink)
             return LocalJob()
 
-        raise RuntimeError(f"Unsupported sink: {sink}.")
+        self._init_spark_session_local_mode()
```
lindong28, line 218: Would it be better to make the code style consistent with `_get_table(...)` by e.g. using `elseif`?

python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +211,41 @@ def materialize_features(
...
+        try:
+            return self._insert_into_spark_supported_sink(
```
lindong28, line 220: It seems unnecessary to specify `supported` here. After all, we do not explicitly check if the sink is supported by spark.

How about `_materialize_dataframe_using_spark`?

python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +211,41 @@ def materialize_features(
...
+        ):
             return self._get_table_from_file_source(features)
         elif isinstance(features, DerivedFeatureView):
             return self._get_table_from_derived_feature_view(features)
         elif isinstance(features, SlidingFeatureView):
             return self._get_table_from_sliding_feature_view(features)
+        elif isinstance(features, FeatureTable):
+            self._init_spark_session_local_mode()
+            try:
+                return self._get_table_from_spark_supported_source(features)
```
lindong28, line 246: How about `_get_table_using_spark`?

python/feathub/processors/local/local_processor.py
```diff
@@ -71,12 +74,21 @@
     append_unix_time_column,
 )
 from feathub.processors.processor import Processor
+from feathub.processors.spark.dataframe_builder.source_sink_utils import (
+    get_dataframe_from_source,
```
lindong28, line 78: It is easy to confuse the method from spark processor with methods in local processor.

How about this:
```
import feathub.processors.spark.dataframe_builder.source_sink_utils as spark_source_sink_utils
```

lindong28 (commented):

python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +210,55 @@ def materialize_features(
...
+        elif is_spark_supported_sink(type(sink)):
+            try:
+                self._init_spark_session_local_mode()
+                return self._materialize_dataframe_using_spark(
+                    df=features_df,
+                    features=features,
+                    sink=sink,
+                    allow_overwrite=allow_overwrite,
+                )
+            except ImportError:
```
lindong28, line 225: Would it be simpler to move this try/catch into `_init_spark_session_local_mode`?

python/feathub/processors/local/local_processor.py
```diff
@@ -178,26 +210,55 @@ def materialize_features(
                 timestamp_field=features.timestamp_field,
                 timestamp_format=features.timestamp_format,
             )
-        elif isinstance(sink, FileSystemSink):
+        elif isinstance(sink, FileSystemSink) and _is_local_file_or_dir(sink.path):
             insert_into_file_sink(features_df, sink)
             return LocalJob()
+        elif is_spark_supported_sink(type(sink)):
```
lindong28, line 216: Would it be simpler to invoke this method as `is_spark_supported_sink(sink)`?

Same for `is_spark_supported_source(...)`.

python/feathub/processors/spark/dataframe_builder/source_sink_utils_common.py
```diff
@@ -0,0 +1,38 @@
...
+from feathub.feature_tables.feature_table import FeatureTable
+from feathub.feature_tables.sinks.black_hole_sink import BlackHoleSink
+from feathub.feature_tables.sinks.file_system_sink import FileSystemSink
+from feathub.feature_tables.sinks.memory_store_sink import MemoryStoreSink
+from feathub.feature_tables.sinks.print_sink import PrintSink
+from feathub.feature_tables.sources.datagen_source import DataGenSource
+from feathub.feature_tables.sources.file_system_source import FileSystemSource
+
+
+def is_spark_supported_source(source_type: Type[FeatureTable]) -> bool:
```
lindong28, line 25: Would it be simpler to add these methods in `source_sink_utils.py`?

lindong28 (commented):

python/feathub/processors/local/local_processor.py
yunfengzhou-hub: `_init_spark_session_local_mode` does not know why it is invoked. If we move this try catch into `_init_spark_session_local_mode`, the error message would be changed from 
> Please install Feathub with Spark to use {sink} in LocalProcessor.

to 

> Please install Feathub with Spark to use certain features like HDFS in LocalProcessor.

Which might cause users not knowing clearly which components require Spark.

python/feathub/processors/spark/dataframe_builder/source_sink_utils_common.py
yunfengzhou-hub: This design aims to ensure users can check whether they can use a connector by installing Spark BEFORE they actually install Spark. If we move these methods to `source_sink_utils.py`, this file would import `source_sink_utils`, and thus importing Spark dependencies. This would cause import errors even if users only need to use local features.

python/feathub/processors/local/local_processor.py
lindong28: We should be able to use the same error message after moving try/catch to `_init_spark_session_local_mode`, right?

python/feathub/processors/spark/dataframe_builder/source_sink_utils_common.py
lindong28: Since we expect this file to only contain methods not depending on pyspark, this means the methods in this file is supposed to be used only by local processor.

Would it be simpler to define these methods in `local_processor.py` similar to `_is_local_file_or_dir`?

python/feathub/processors/local/local_processor.py
yunfengzhou-hub: Yes, it would simplify the code a bit, but it also makes the error message not show the concrete component that requires Spark.

lindong28: How about moving both try/catch and `_init_spark_session_local_mode` into `_materialize_dataframe_using_spark`?

Same for `_get_table_using_spark`.

lindong28:
Thanks for the update. LGTM.

PR #127, opened 2023-04-26
FlinkProcessor FileSystem/Kakfa source/sink support protobuf format

Sxnan:

This PR update FlinkProcessor to support protobuf format for FileSystem and Kafka source/sink.

## Brief change log

- Update FeatureTable to allow setting configuration for the data format.
- Update FlinkProcessor FileSystem/Kakfa source/sink support protobuf format.

  - The public API: yes

## Documentation

### DIFF
--- a/docs/content/connectors/formats/json.md
+++ b/docs/content/connectors/formats/json.md
@@ -1,0 +1,27 @@
+# JSON Format
+
+The JSON format allows to read and write JSON data based on an JSON schema. The JSON 
+schema is derived from table schema.
+
+## Format Options
+
+| key                | Required | default | type    | Description                                                                                          |
+|--------------------|----------|---------|---------|------------------------------------------------------------------------------------------------------|
+| ignore_parse_error | Optional | True    | boolean | Skip fields and rows with parse errors instead of failing. Fields are set to null in case of errors. |
+
+## Data Type Mapping
+
+The following table lists the type mapping from Feathub type to JSON type.
+
+| Feathub type | JSON type                     |
+|--------------|-------------------------------|
+| Bytes        | string with encoding: base64  |
+| String       | string                        |
+| Int32        | number                        |
+| Int64        | number                        |
+| Float32      | number                        |
+| Float64      | number                        |
+| Bool         | boolean                       |
+| Timestamp    | string with format: date-time |
+| VectorType   | array                         | 
+| MapType      | object                        | 
--- a/docs/content/connectors/formats/protobuf.md
+++ b/docs/content/connectors/formats/protobuf.md
@@ -1,0 +1,97 @@
+# Protobuf Format
+
+The Protocol Buffers [Protobuf](https://developers.google.com/protocol-buffers) format 
+allows you to read and write Protobuf data, based on Protobuf generated classes.
+
+## Format Options
+
+| key                 | Required | default | type    | Description                                                                                                                                                                                       |
+|---------------------|----------|---------|---------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| protobuf.jar_path   | Required | (none)  | string  | The path to the jar that contains the Protobuf generated classes.                                                                                                                                 |
+| protobuf.class_name | Required | (none)  | string  | The full name of a Protobuf generated class. The name must match the message name in the proto definition file. $ is supported for inner class names, like 'com.exmample.OuterClass$MessageClass' |
+| ignore_parse_error  | Optional | True    | boolean | Skip fields and rows with parse errors instead of failing. Fields are set to null in case of errors.                                                                                              |
+
+## How to create a table with Protobuf format
+
+Here is an example to create a table using the Kafka connector and Protobuf format.
+
+Below is the proto definition file.
+
+```protobuf
+syntax = "proto3";
+
+package org.feathub.proto;
+option java_package = "org.feathub.proto";
+option java_multiple_files = true;
+
+message KafkaTestMessage {
+  int64 id = 1;
+  int64 val = 2;
+  string ts = 3;
+}
+```
+
+1. Use [protoc](https://developers.google.com/protocol-buffers/docs/javatutorial#compiling-your-protocol-buffers)
+   command to compile the .proto file to java classes.
+2. Compile and package the classes. 
+   [feathub-test-proto](../../../../java/feathub-test-proto) is an example that use 
+   maven to compile and package the .proto file.
+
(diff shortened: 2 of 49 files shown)


Sxnan:

docs/content/connectors/formats/json.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Data Type Mapping
+
+The following table lists the type mapping from Flink type to JSON type.
+
+| Feathub type | JSON type                     |
+|--------------|-------------------------------|
+| BYTES        | string with encoding: base64  |
+| STRING       | string                        |
+| INT32        | number                        |
```
lindong28, line 20: It seems inconsistent to use int32, int64, float and double.

How about we rename INT32 as INT and rename `INT64` as Long, for consistency with Spark DataType [1]?

Flink uses `BIGINT` to denote "Data type of an 8-byte signed integer". However, Java uses BigInteger to represent arbitrary-precision integers [2]. Thus it seems better to use long.

[1] https://spark.apache.org/docs/latest/sql-ref-datatypes.html
[2] https://www.baeldung.com/java-biginteger

java/feathub-test-proto/src/proto/message.proto
```diff
@@ -0,0 +1,22 @@
...
+option java_multiple_files = true;
+
+message FileSystemTestMessage {
+  // TODO: We have to add a dummy field with type int64 at the beginning,
+  //  otherwise exception will be thrown.
+  //  https://issues.apache.org/jira/browse/FLINK-31944
+  int64 dummy = 1;
+  string name = 2;
+  int64 cost = 3;
+  int64 distance = 4;
```
lindong28, line 14: Can we have a test that covers all supported types? Maybe just add fields in this message and verify that the fields can be written and read correctly?

python/feathub/feature_tables/tests/test_file_system_source_sink.py
```diff
@@ -42,10 +48,28 @@ def test_local_file_system_json_source_sink(self):
...
+    def test_local_file_system_protobuf_source_sink(self):
+        sink_path = tempfile.NamedTemporaryFile(
+            dir=self.temp_dir, suffix=".protobuf"
+        ).name
+        self._test_file_system_source_sink(
+            sink_path,
+            "protobuf",
+            {
+                PROTOBUF_JAR_PATH_CONFIG: get_protobuf_jar_path(),
+                PROTOBUF_CLASS_NAME_CONFIG: "org.feathub.proto.FileSystemTestMessage",
```
lindong28, line 60: `message.proto` is currently placed in "java/feathub-test-proto/src/proto" and its compilation depends on the code specified in ./java/feathub-test-proto/pom.xml.

It will be great if we can provide a more pythonic end-to-end solution that allows user to specify a protobuf file in python space with built-in infra to compile and use the protobuf in FeatHub. Is it feasible? I am wondering how pyflink or pyspark allows user to use protobuf.

BTW, it might be useful to checkout how `python/setup.py` compiles `python/feathub/common/protobuf/value.proto` with `generate_py_protobuf`, and re-use as much infra as possible.

python/feathub/feature_tables/sinks/kafka_sink.py
```diff
@@ -38,17 +40,26 @@ def __init__(
...
+                                  message key.
+        :param value_format_config: Optional. The configuration of the format for Kafka
+                                    message value.
         """
         super().__init__(
             name="",
             system_name="kafka",
             properties={"bootstrap_server": bootstrap_server, "topic": topic},
             data_format=value_format,
+            format_config=value_format_config,
```
lindong28, line 53: The `Sink#__init__(...)` does not have this named parameter.

And it might be simpler to encode extra configs in `properties` in a pattern similar to Kafka Source [1].

[1] https://github.com/alibaba/feathub/blob/master/python/feathub/processors/flink/table_builder/tests/test_kafka_source_sink.py#L116

python/feathub/feature_tables/sinks/kafka_sink.py
```diff
@@ -38,17 +40,26 @@ def __init__(
                              etc.
         :param producer_properties: Optional. If it is not None, it contains the extra
                                     kafka producer properties.
+        :param key_format_config: Optional. The configuration of the format for Kafka
```
lindong28, line 43: Would it be better to name it `key_format_properties` for consistency with `producer_properties`?

We can simplify `properties` to `props` if you prefer.

python/feathub/feature_tables/tests/test_kafka_source_sink.py
```diff
@@ -65,7 +71,24 @@ def tearDownClass(cls) -> None:
         super().tearDownClass()
         cls.kafka_container.stop()
 
-    def test_kafka_source_sink(self):
+    def test_kafka_source_sink_json(self):
+        self._test_kafka_source_sink("json")
```
lindong28, line 75: It will be great to have the test mimic users' experience by connecting the protobuf definition with its usage in the test.

After all, we expect there is an easy way for users to define and use protobuf in FeatHub.

lindong28 (commented):

docs/content/connectors/formats/json.md
Sxnan: I find out that the type names in the docs were inaccurate.  The naming of the Feathub type is consistent, int32, int64, float32, and float64.

I updated the docs so they are consistent now.

Sxnan:

docs/content/connectors/formats/protobuf.md
```diff
@@ -0,0 +1,97 @@
...
+    key_format=None,
+    value_format="protobuf",
+    schema=schema,
+    consumer_group="test-group",
+    keys=["id"],
+    timestamp_field="ts",
+    timestamp_format="%Y-%m-%d %H:%M:%S",
+    value_format_config={
+        "protobuf.jar_path": "jar_path",  # The path to the jar packaged above.
+       "protobuf.class_name": "org.feathub.proto.KafkaTestMessage"
```
lindong28, line 64: nits: use the same indention as the line below.

lindong28 (commented):
Thanks for the update! LGTM overall. Left some minor comments.

lindong28 (commented):
Thanks for the update! LGTM overall. Left some minor comments.

lindong28 (commented):
Thanks for the update! LGTM overall. Left some minor comments.

lindong28:
Thanks for the update. LGTM.

PR #133, opened 2023-05-11
PythonUdfTransform supports default value on exception

Sxnan:

This PR update the `PythonUdfTransform` to support setting a default value when an exception is raised. 

## Brief change log

- PythonUdfTransform supports default value on exception

  - The public API:yes

## Documentation

### DIFF
--- a/python/feathub/feature_views/transforms/python_udf_transform.py
+++ b/python/feathub/feature_views/transforms/python_udf_transform.py
@@ -24,14 +24,52 @@ class PythonUdfTransform(Transformation):
     time.
     """
 
-    def __init__(self, udf: Callable[[pd.Series], Any]) -> None:
+    # TODO: Validate the type of default value with feature type.
+    def __init__(
+        self,
+        udf: Callable[[pd.Series], Any],
+        fail_on_exception: bool = True,
+        value_on_exception: Any = None,
+    ) -> None:
         """
         :param udf: The udf that will be invoked for each row. The input
                           of the udf is a Pandas Series object that represent the
                           row.
+        :param fail_on_exception: Whether to fail the job in case of an exception is
+                                  raised by the udf. If this is false, the
+                                  `value_on_exception` is used in case of exception.
+        :param value_on_exception: If `fail_on_exception` is set to false, this is the
+                              default value of the result when an exception is raised.
         """
         super().__init__()
-        self.udf = udf
+        self.original_udf = udf
+        self.fail_on_exception = fail_on_exception
+        self.value_on_exception = value_on_exception
+
+    @property
+    def udf(self) -> Callable[[pd.Series], Any]:
+        if self.fail_on_exception:
+            return self.original_udf
+
+        return self._wrap_udf_with_value_on_exception(
+            self.original_udf, self.value_on_exception
+        )
+
+    @staticmethod
+    def _wrap_udf_with_value_on_exception(
+        udf: Callable[[pd.Series], Any], default_value: Any
+    ) -> Callable[[pd.Series], Any]:
+        def wrapper(s: pd.Series) -> Any:
+            try:
+                return udf(s)
+            except Exception:
+                return default_value
+
+        return wrapper
 
     def to_json(self) -> Dict:
-        return {"type": "PythonTransform"}
+        return {
+            "type": "PythonUdfTransform",
+            "fail_on_exception": self.fail_on_exception,
+            "value_on_exception": self.value_on_exception,
+        }
--- a/python/feathub/feature_views/transforms/tests/test_python_udf_transform.py
+++ b/python/feathub/feature_views/transforms/tests/test_python_udf_transform.py
@@ -99,3 +99,45 @@ class PythonUDFTransformITTest(ABC, FeathubITTestBase):
         )
 
         assert_frame_equal(expected_result_df, result_df)
+
+    def test_python_udf_with_default_value(self):
+        df_1 = self.input_data.copy()
+        source = self.create_file_source(df_1)
+
+        def raise_on_alex(row: pd.Series) -> str:
+            if row["name"] == "Alex":
+                raise RuntimeError()
+            return row["name"]
+
+        feature_view = DerivedFeatureView(
+            name="feature_view",
+            source=source,
+            features=[
+                Feature(
+                    name="new_name",
+                    dtype=String,
+                    transform=PythonUdfTransform(
+                        raise_on_alex,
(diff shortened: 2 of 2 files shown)


Sxnan:

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,48 @@ class PythonUdfTransform(Transformation):
     time.
     """
 
-    def __init__(self, udf: Callable[[pd.Series], Any]) -> None:
+    def __init__(
+        self,
+        udf: Callable[[pd.Series], Any],
+        fail_on_exception: bool = True,
```
yunfengzhou-hub, line 30: A similar configuration I've seen is `HasHandleInvalid` in the Flink ML project, which has the following strategies when dealing with invalid values or errors:
- `error`: throws out the internal exception
- `keep`: outputs a default value
- `skip`: filters out such rows and do not output them

Do you think we should add an option that corresponds to the `skip` strategy in Feathub?

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,48 @@ class PythonUdfTransform(Transformation):
...
+        default_value: Any = None,
+    ) -> None:
         """
         :param udf: The udf that will be invoked for each row. The input
                           of the udf is a Pandas Series object that represent the
                           row.
+        :param fail_on_exception: Whether to fail the job in case of an exception is
+                                  raised by the udf.
+        :param default_value: If `fail_on_exception` is set to false, this is the
+                              default value of the result when an exception is raised.
```
yunfengzhou-hub, line 40: Would it be better if Feathub checks whether the default value meets its data type in schema? For example, Feathub may throw exception in the following situations:
- data type is boolean, but the default value is None
- data type is integer, but the default value is "foobar".

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,48 @@ class PythonUdfTransform(Transformation):
...
+        self,
+        udf: Callable[[pd.Series], Any],
+        fail_on_exception: bool = True,
+        default_value: Any = None,
+    ) -> None:
         """
         :param udf: The udf that will be invoked for each row. The input
                           of the udf is a Pandas Series object that represent the
                           row.
+        :param fail_on_exception: Whether to fail the job in case of an exception is
```
yunfengzhou-hub, line 37: It might be better to denote that the job would output a default value in case of exceptions here or in the document of this class, instead of in the document for the `default_value` parameter.

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,48 @@ class PythonUdfTransform(Transformation):
...
+                return udf(s)
+            except Exception:
+                return default_value
+
+        return wrapper
 
     def to_json(self) -> Dict:
-        return {"type": "PythonTransform"}
+        return {
+            "type": "PythonTransform",
```
yunfengzhou-hub, line 68: nit: Let's update this to `type: PythonUdfTransform`.

yunfengzhou-hub (commented):

python/feathub/feature_views/transforms/python_udf_transform.py
Sxnan: I think `skip` strategy may not be a good strategy to use in Feathub. In Feathub, user defines the computation logic of each feature, and each feature is a column in the row. I think skipping the entire row when one of the features is invalid is undesirable.  

I think it makes sense in FlinkML, because in FlinkML the input is a row and the output is also a row.

python/feathub/feature_views/transforms/python_udf_transform.py
Sxnan: Currently, if the default value is incompatible with the data type of the feature, the exception will be thrown at runtime by the processor. It is a good improvement to validate in Feathub and fail earlier. The improvement is not trivial so I added a TODO. What do you think?

Sxnan:
@yunfengzhou-hub Thanks for the review. I updated the PR accordingly. Can you have a another look?

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,50 @@ class PythonUdfTransform(Transformation):
...
+        fail_on_exception: bool = True,
+        default_value: Any = None,
+    ) -> None:
         """
         :param udf: The udf that will be invoked for each row. The input
                           of the udf is a Pandas Series object that represent the
                           row.
+        :param fail_on_exception: Whether to fail the job in case of an exception is
+                                  raised by the udf. If this is false, the
+                                  `default_value` is used  in case of exception.
```
yunfengzhou-hub, line 40: nit: there are two whitespaces between "used" and "in".

yunfengzhou-hub (commented):
Thanks for the update. LGTM except the following minor comment. @lindong28 Could you please take a look at this PR?

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,50 @@ class PythonUdfTransform(Transformation):
     time.
     """
 
-    def __init__(self, udf: Callable[[pd.Series], Any]) -> None:
+    # TODO: Validate the type of default value with feature type.
+    def __init__(
+        self,
+        udf: Callable[[pd.Series], Any],
+        fail_on_exception: bool = True,
+        default_value: Any = None,
```
lindong28, line 32: Would it be more intuitive to name it `value_on_exception`?

python/feathub/feature_views/transforms/python_udf_transform.py
```diff
@@ -24,14 +24,50 @@ class PythonUdfTransform(Transformation):
...
                           row.
+        :param fail_on_exception: Whether to fail the job in case of an exception is
+                                  raised by the udf. If this is false, the
+                                  `default_value` is used  in case of exception.
+        :param default_value: If `fail_on_exception` is set to false, this is the
+                              default value of the result when an exception is raised.
         """
         super().__init__()
-        self.udf = udf
+        self._udf = udf
```
lindong28, line 45: How about naming it `original_udf` so that the name does not start with `_`?

lindong28 (commented):
Thanks for the PR. Left two minor comments.

lindong28:
Thanks for the update! LGTM.

PR #134, opened 2023-05-11
Support user defined attributes in Feature and FeatureView

Sxnan:

This PR updates the feature and feature views to allow user-defined attributes.

## Brief change log

- Support user defined attributes in Feature and FeatureView

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/feature_views/feature.py
+++ b/python/feathub/feature_views/feature.py
@@ -39,6 +39,8 @@ class Feature:
         dtype: Optional[DType] = None,
         keys: Optional[Sequence[str]] = None,
         input_features: Sequence[Feature] = (),
+        description: str = "",
+        extra_props: Optional[Dict[str, str]] = None,
     ):
         """
         :param name: The name that uniquely identifies this feature in the
@@ -54,6 +56,9 @@ class Feature:
                      the `transform` or the parent table's keys.
         :param input_features: The names of fields in the parent table used by
                               `transform` to derive this feature's values.
+        :param description: The description of the feature.
+        :param extra_props: The extra properties of the feature that are defined by
+                            user.
         """
         if name.startswith("__") or name.endswith("__"):
             raise FeathubException(
@@ -81,6 +86,8 @@ class Feature:
             keys = sorted(keys)
         self.keys = keys
         self.input_features = input_features
+        self.description = description
+        self.extra_props = {} if extra_props is None else extra_props
 
     def to_json(self) -> Dict:
         return {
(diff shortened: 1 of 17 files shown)


Sxnan:

python/feathub/feature_views/feature.py
```diff
@@ -54,6 +55,7 @@ def __init__(
                      the `transform` or the parent table's keys.
         :param input_features: The names of fields in the parent table used by
                               `transform` to derive this feature's values.
+        :param attributes: The extra attribute of the feature that are defined by user.
```
yunfengzhou-hub, line 58: nit: change `The extra attribute` to `The extra attributes`. Same for other classes.

python/feathub/table/table_descriptor.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         keys: Optional[List[str]],
         timestamp_field: Optional[str],
         timestamp_format: str = "epoch",
+        attributes: Optional[Dict[str, str]] = None,
```
yunfengzhou-hub, line 57: Would it be better to reserve some keywords that might be applicable to all tables? For example, reserve all keywords that start and end with double underscores, so that Feathub may define common attributes like `__DESCRIPTION__`.

yunfengzhou-hub (commented):

python/feathub/table/table_descriptor.py
Sxnan: I think Feathub doesn't need to reserve keywords in the attribute and users can always add the description to the attributes, e.g. "description": "...". What do you think?

Sxnan:
@yunfengzhou-hub  Thanks for the review. I updated the PR accordingly. Can you have another look?

yunfengzhou-hub (commented):
Thanks for the update. LGTM. @lindong28 Could you please take a look at this PR?

python/feathub/table/table_descriptor.py
lindong28, line 57: Would it be better to name it `extra_props` for consistency with `FlinkProcessor.props` and `FeatureTable.data_format_properties`?

python/feathub/feature_views/feature_view.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         keep_source_fields: bool = False,
         timestamp_field: Optional[str] = None,
         timestamp_format: str = "epoch",
+        attributes: Optional[Dict[str, str]] = None,
```
lindong28, line 48: Do we have a use-case for specifying the per-feature-view attribute in addition to the per-feature attribute?

python/feathub/table/table_descriptor.py
lindong28: We need to predefine field names for common information such as `description`.  This is necessary for UI to programmatically analyze and display that common information.

Maybe add `description: str` to the `Feature` constructor?

lindong28 (commented):

lindong28 (commented):

lindong28:
Thanks for the update. LGTM.

PR #135, opened 2023-05-15
Support Redis data types in Redis connector

yunfengzhou-hub:

This pull request adds support for Redis data types instead of protobuf-serialized bytes in Redis connector.

## Brief change log

- Changes the format of feature values saved to Redis, using Redis data types.
- Removes value updates with timestamp comparison.

  - The public API: yes

## Documentation


java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/sink/RedisSinkFunction.java
```diff
@@ -48,86 +56,77 @@
...
+                        getList(data.getArray(i), (CollectionDataType) dataType)
+                                .toArray(new String[0]));
+            } else {
+                client.set(key, getString(data, i, dataType));
             }
-
-            this.client.evalsha(this.evalScriptSHA, Collections.singletonList(key), arguments);
         }
+
+        client.flush();
```
lindong28, line 129: `SinkFunction#invoke` does not have the flush semantics. Would it be better to execute `sync()` only in `SinkFunction#finish`?

lindong28 (commented):

lindong28 (commented):

java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/sink/JedisMasterClient.java
```diff
@@ -38,25 +40,55 @@ public JedisMasterClient(String host, int port, String username, String password
...
+        if (pipeline == null) {
+            pipeline = jedis.pipelined();
+        }
+        pipeline.set(key, value);
+    }
+
+    @Override
+    public void flush() {
+        if (pipeline != null) {
+            pipeline.sync();
```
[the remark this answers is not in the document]
yunfengzhou-hub: The JavaDoc of `redis.clients.jedis.Pipeline#sync` shows the following contents.
```java
  /**
   * Synchronize pipeline by reading all responses. This operation close the pipeline. In order to
   * get return values from pipelined commands, capture the different Response&lt;?&gt; of the
   * commands you execute.
   */
```
As this operation closes the pipeline, I supposed that we need to re-instantiate the pipeline then.

However, I found that Feathub is still able to write out content to Redis using the same pipeline after it is synced and closed, so it might be better to reuse one pipeline for writing multiple rows of features.

java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/sink/RedisSinkFunction.java
yunfengzhou-hub: Invoking `flush()` only at `SinkFunction#finish` can work only when the input stream is bounded. As it has been avoided to re-instantiate the pipeline for every record as described in the first comment, I think flushing out data at `SinkFunction#invoke` would not have that much performance overhead. What do you think?

lindong28: Calling flush() on every record typically will reduce throughput because there might be much less data sent in each round-trip time. After all, there is no need to provide an API to explicitly flush data if there is no latency/throughput tradeoff.

Flink is typically used for use-case with nearline latency with high throughput requirement. So it seems better not to call flush() for every record. We need to call flush() as part of close().

java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/sink/JedisMasterClient.java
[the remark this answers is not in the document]
lindong28: Sounds good.

lindong28:
Thanks for the update. LGTM.

PR #136, opened 2023-05-18
Add support for MySqlRegistry

yunfengzhou-hub:

This pull request mainly adds MySqlRegistry so that Feathub feature metadata can be persisted remotely and synced across a distributed environment.

## Brief change log

- Adds MySqlRegistry.
- Adds from_json() method to TableDescriptor and relevant classes

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/feathub_client.py
+++ b/python/feathub/feathub_client.py
@@ -17,6 +17,7 @@ import pandas as pd
 from datetime import datetime, timedelta
 
 from feathub.common.config import flatten_dict
+from feathub.common.utils import deprecated_alias
 from feathub.feature_tables.feature_table import FeatureTable
 from feathub.processors.processor import Processor
 from feathub.registries.registry import Registry
@@ -44,9 +45,10 @@ class FeathubClient:
             registry=self.registry,
         )
 
+    @deprecated_alias(features="feature_descriptor")
     def get_features(
         self,
-        features: Union[str, TableDescriptor],
+        feature_descriptor: Union[str, TableDescriptor],
         keys: Union[pd.DataFrame, TableDescriptor, None] = None,
         start_datetime: Optional[datetime] = None,
         end_datetime: Optional[datetime] = None,
@@ -54,9 +56,9 @@ class FeathubClient:
         """
         Returns a table of features according to the specified criteria.
 
-        :param features: Describes the features to be included in the table. If it is a
-                         string, it refers to the name of a table descriptor in the
-                         entity registry.
+        :param feature_descriptor: Describes the features to be included in the table.
+                                   If it is a string, it refers to the name of a table
+                                   descriptor in the entity registry.
         :param keys: Optional. If it is TableDescriptor or DataFrame, it should be
                      transformed into a table of keys. If it is not None, the output
                      table should only include features whose key fields match a row of
@@ -75,15 +77,16 @@ class FeathubClient:
         :return: A table of features.
         """
         return self.processor.get_table(
-            features=features,
+            feature_descriptor=feature_descriptor,
             keys=keys,
             start_datetime=start_datetime,
             end_datetime=end_datetime,
         )
 
+    @deprecated_alias(features="feature_descriptor")
     def materialize_features(
         self,
-        features: Union[str, TableDescriptor],
+        feature_descriptor: Union[str, TableDescriptor],
         sink: FeatureTable,
         ttl: Optional[timedelta] = None,
         start_datetime: Optional[datetime] = None,
@@ -94,9 +97,9 @@ class FeathubClient:
         Starts a job to write a table of features into the given sink according to the
         specified criteria.
 
-        :param features: Describes the table of features to be inserted in the sink. If
-                         it is a string, it refers to the name of a table descriptor in
-                         the entity registry.
+        :param feature_descriptor: Describes the table of features to be inserted in the
+                                   sink. If it is a string, it refers to the name of a
+                                   table descriptor in the entity registry.
         :param sink: Describes the location to write the features.
         :param ttl: Optional. If it is not None, the features data should be purged from
                     the sink after the specified period of time.
@@ -111,7 +114,7 @@ class FeathubClient:
         :return: A processor job corresponding to this materialization operation.
         """
         return self.processor.materialize_features(
-            features=features,
+            feature_descriptor=feature_descriptor,
             sink=sink,
             ttl=ttl,
             start_datetime=start_datetime,
@@ -137,8 +140,12 @@ class FeathubClient:
             feature_names=feature_names,
         )
 
+    @deprecated_alias(features_list="feature_descriptors")
     def build_features(
-        self, features_list: List[TableDescriptor], props: Optional[Dict] = None
+        self,
+        feature_descriptors: List[TableDescriptor],
+        force_update: bool = False,
+        props: Optional[Dict] = None,
     ) -> List[TableDescriptor]:
         """
         For each table descriptor in the given list, resolve this descriptor by
@@ -151,9 +158,35 @@ class FeathubClient:
         And caches the resolved table descriptors in memory so that they can be used
         when building other table descriptors.
 
-        :param features_list: A list of table descriptors.
+        :param feature_descriptors: A list of table descriptors.
+        :param force_update: If True, the feature descriptor would be directly searched
+                             in registry. If False, the feature descriptor would be
+                             searched in local cache first.
         :param props: Optional. If it is not None, it is the global properties that are
                       used to configure the given table descriptors.
         :return: A list of resolved descriptors corresponding to the input descriptors.
         """
-        return self.registry.build_features(features_list=features_list, props=props)
+        return self.registry.build_features(
+            feature_descriptors=feature_descriptors,
+            force_update=force_update,
+            props=props,
+        )
+
+    def register_features(
+        self, feature_descriptors: List[TableDescriptor], force_update: bool = False
+    ) -> List[bool]:
+        """
+        Registers the given table descriptor in the registry after building and
+        caching them in memory as described in build_features. Each descriptor is
+        uniquely identified by its name in the registry.
+
+        :param feature_descriptors: A table descriptor to be registered.
+        :param force_update: If True, the feature descriptor would be directly searched
+                             in registry. If False, the feature descriptor would be
+                             searched in local cache first.
+        :return: A list of bool values denoting whether the registration for each
+                 descriptor is successful.
+        """
+        return self.registry.register_features(
+            feature_descriptors=feature_descriptors, force_update=force_update
+        )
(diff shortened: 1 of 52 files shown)


python/feathub/registries/registry.py
```diff
@@ -111,6 +121,11 @@ def instantiate(props: Dict) -> Registry:
 
             return LocalRegistry(props=props)
 
+        elif registry_type == RegistryType.MYSQL:
```
lindong28, line 124: Can we remove the empty line above?

lindong28 (commented):

lindong28 (commented):

python/feathub/registries/local_registry.py
```diff
@@ -80,16 +80,18 @@ def register_features(
             raise FeathubException("Cannot register a TableDescriptor with empty name.")
         if features.name in self.tables and not override:
             return False
-        self.tables[features.name] = features
+        self.tables[features.name] = (features, features.build(self, None))
```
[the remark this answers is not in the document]
yunfengzhou-hub: The same descriptor might be resolved into different features if its sources are updated, so it is necessary to support rebuilding a feature descriptor.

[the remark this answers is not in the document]
lindong28: Suppose users carry out the following sequence of steps, the FeatureView they register should be the same as the FeatureView they have validated, right?

```
- Specify FeatureView
- Build FeatureView
- Validate and test FeatureView
- Register the validated FeatureView
```

python/feathub/registries/local_registry.py
```diff
@@ -55,41 +55,46 @@ def __init__(self, props: Dict) -> None:
...
+                raise FeathubException(
+                    "Cannot register a TableDescriptor with empty name."
+                )
+            if descriptor.name in self.tables and not override:
+                return False
+            self.tables[descriptor.name] = (descriptor, descriptor.build(self, None))
         return True
 
-    def get_features(self, name: str) -> TableDescriptor:
+    def get_features(
```
lindong28, line 89: Can we also add `force_update: bool = False` to `build_features(...)`?

python/feathub/feathub_client.py
```diff
@@ -46,17 +46,17 @@ def __init__(self, props: Dict) -> None:
 
     def get_features(
         self,
-        features: Union[str, TableDescriptor],
+        feature_descriptor: Union[str, TableDescriptor],
```
lindong28, line 49: Can we follow the example provided in this thread [1] to rename the argument in a backward-compatible way?

[1] https://stackoverflow.com/questions/49802412/how-to-implement-deprecation-in-python-with-argument-alias

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+        self.cursor.execute(
+            f"""
+                CREATE TABLE IF NOT EXISTS `{self.table}`(
+                   `digest` VARCHAR(64) NOT NULL,
+                   `name` TEXT NOT NULL,
+                   `timestamp` TIMESTAMP NOT NULL,
+                   `is_deleted` BOOLEAN NOT NULL,
+                   `original_json` TEXT NOT NULL,
+                   `built_json` TEXT NOT NULL,
+                   PRIMARY KEY ( `digest`, `timestamp` )
```
lindong28, line 138: How about using `name` and `timestamp` as the primary keys?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+            )
+
+        self.tables[name] = existing_feature
+        return self.tables[name][1 if is_built else 0]
+
+    def delete_features(self, name: str) -> bool:
+        self.cursor.execute(
+            f"""
+                UPDATE {self.table}
+                SET `is_deleted`=True
```
lindong28, line 228: I am not sure we need `is_delete` field.

Would it be simpler to just delete the corresponding row from mysql?

python/feathub/registries/registry.py
```diff
@@ -63,26 +71,35 @@ def build_features(
...
+                             in registry. If False, the feature descriptor would be
+                             searched in local cache first.
         :return: True iff registration is successful.
         """
         pass
 
     @abstractmethod
-    def get_features(self, name: str) -> TableDescriptor:
+    def get_features(
+        self, name: str, force_update: bool = False, is_built: bool = True
```
lindong28, line 91: It might be more intuitive to name the field `is_resolved`.

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+    ) -> bool:
+        self.build_features(feature_descriptors, force_update=force_update)
+
+        for descriptor in feature_descriptors:
+            existing_features = self._get_feature_from_remote_registry_or_none(
+                descriptor.name
+            )
+            if (
+                existing_features is not None
+                and existing_features[1].to_json()
```
lindong28, line 174: Would it be more intuitive to compare the digest?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+            ):
+                continue
+
+            original_json = json.dumps(
+                self.tables[descriptor.name][0].to_json(), sort_keys=True
+            ).replace('"', '\\"')
+            built_json = json.dumps(
+                self.tables[descriptor.name][1].to_json(), sort_keys=True
+            )
+            digest = _get_digest(built_json)
```
lindong28, line 185: Should digest also depend on the `original_descriptor`?

And would it be simpler to calculate digest after `built_json.replace('"', '\\"')`?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+
+        return result
+
+    def register_features(
+        self, feature_descriptors: List[TableDescriptor], force_update: bool = False
+    ) -> bool:
+        self.build_features(feature_descriptors, force_update=force_update)
+
+        for descriptor in feature_descriptors:
+            existing_features = self._get_feature_from_remote_registry_or_none(
```
lindong28, line 169: `build_features(` invoked above should have made sure that the given list of `feature_descriptors` can be found in `self.tables`. Would it be simpler to read `self.tables` directly without invoking `_get_feature_from_remote_registry_or_none()` here?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+            return self.tables[name][1 if is_built else 0]
+
+        existing_feature = self._get_feature_from_remote_registry_or_none(name)
+        if existing_feature is None:
+            raise RuntimeError(
+                f"Table '{name}' is not found in the cache or registry. "
+                "Please invoke build_features(..) for this table."
+            )
+
+        self.tables[name] = existing_feature
```
lindong28, line 221: Would it be more intuitive to update `self.tables()` in `_get_feature_from_remote_registry_or_none()`?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,273 @@
...
+                SET `is_deleted`=True
+                WHERE `name` = "{name}"
+                ORDER BY `timestamp` DESC
+                LIMIT 1;
+            """
+        )
+        self.tables.pop(name)
+        return True
+
+    def _get_feature_from_remote_registry_or_none(
```
lindong28, line 237: Would it be simpler to name it `_get_descriptor_from_mysql`?

lindong28 (commented):

lindong28 (commented):

python/feathub/registries/mysql_registry.py
yunfengzhou-hub: MySQL requires that columns that are used as keys must have a limited length, while Feathub does not have this requirement for `name` strings.

yunfengzhou-hub: MySQL requires that key columns have a fixed length, while Feathub has not yet limited the size of feature descriptor names. I have required that descriptors to be registered into MySQL must have their names less than 64 characters, and added a TODO that we may apply this requirement to all Features and TableDescriptors in the future.

lindong28: Sounds good.

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,289 @@
...
+                self.tables[descriptor.name][1].to_json(), sort_keys=True
+            )
+            digest = _get_digest(original_descriptor, resolved_descriptor)
+
+            if existing_descriptor is not None and existing_descriptor[2] == digest:
+                continue
+
+            original_descriptor = original_descriptor.replace('"', '\\"')
+            resolved_descriptor = resolved_descriptor.replace('"', '\\"')
+            timestamp = self.tables[descriptor.name][2].strftime("%Y-%m-%d %H:%M:%S")
```
lindong28, line 200: Can we also encode the timezone info?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,289 @@
...
+            self.cursor.execute(
```
lindong28, line 201: Can we compare the timestamp and only insert an entry if the timestamp in the mysql is older?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,289 @@
...
+                table.build(self, force_update=force_update, props=props),
+                datetime.utcnow(),
+            )
+            result.append(self.tables[table.name][1])
+
+        return result
+
+    def register_features(
+        self, feature_descriptors: List[TableDescriptor], force_update: bool = False
+    ) -> bool:
```
lindong28, line 181: Can you update this method to return `List[bool]` so that the caller knows which descriptors are successfully registered?

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,289 @@
...
+                        "{resolved_descriptor}"
+                    );
+                """
+            )
+        return True
+
+    def get_features(
+        self, name: str, force_update: bool = False, is_resolved: bool = True
+    ) -> TableDescriptor:
+        if not force_update and name in self.tables:
```
lindong28, line 223: The following code might be more readable:

```
if force_update:
    self._get_descriptor_from_mysql(name)
if name not in self.tables:
    raise RuntimeError(
        f"Table '{name}' is not found in the cache or registry. "
        "Please invoke build_features(..) for this table."
    )
return self.tables[name][1 if is_resolved else 0]
```

python/feathub/registries/mysql_registry.py
```diff
@@ -0,0 +1,289 @@
...
+
+    def delete_features(self, name: str) -> bool:
+        self.cursor.execute(
+            f"""
+                DELETE FROM {self.table}
+                WHERE `name` = "{name}";
+            """
+        )
+        self.tables.pop(name)
+        return True
```
lindong28, line 249: Can we return `False` if the feature is not found in the registry?

lindong28 (commented):

lindong28 (commented):

lindong28:
Thanks for the update! LGTM.

PR #137, opened 2023-05-19
Add from_json to TableDescriptors

yunfengzhou-hub:

This pull request adds from_json method to TableDescriptor and relevant classes.

## Brief change log

- Adds from_json method to TableDescriptor and relevant classes.

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/common/types.py
+++ b/python/feathub/common/types.py
@@ -12,14 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from enum import Enum
+import json
 from abc import ABC, abstractmethod
+from enum import Enum
 from typing import Type, Dict
 
 import numpy as np
-import json
 
-from feathub.common.exceptions import FeathubTypeException, FeathubExpressionException
+from feathub.common.exceptions import (
+    FeathubTypeException,
+    FeathubExpressionException,
+)
+from feathub.common.utils import append_metadata_to_json, from_json
 
 
 class BasicDType(Enum):
@@ -62,8 +66,13 @@ class PrimitiveType(DType):
         super().__init__()
         self.basic_dtype = basic_dtype
 
+    @append_metadata_to_json
     def to_json(self) -> Dict:
-        return {"type": "PrimitiveType", "basic_dtype": f"{self.basic_dtype.name}"}
+        return {"basic_dtype": f"{self.basic_dtype.name}"}
+
+    @classmethod
+    def from_json(cls, json_dict: Dict) -> "PrimitiveType":
+        return PrimitiveType(BasicDType[json_dict["basic_dtype"]])
 
 
 class VectorType(DType):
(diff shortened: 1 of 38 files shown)


python/feathub/common/types.py
```diff
@@ -38,7 +42,10 @@ class BasicDType(Enum):
 
 class DType(ABC):
     def __init__(self) -> None:
-        pass
+        self.to_json = append_metadata_to_json(  # type: ignore
```
lindong28, line 45: It is not easy to know whether `to_json` is decorated here by reading the code.

To make the code more readable, can we either use the python decorator by annotating the `to_json()` with `@append_metadata_to_json()`, or add the metadata (e.g. version, type) directly in the original `to_json()`?

Note that we can use a constant global variable to specify the current json format version.

python/feathub/processors/flink/job_submitter/feathub_job_descriptor.py
```diff
@@ -69,6 +70,57 @@ def __init__(
         self.local_registry_tables = local_registry_tables
         self.allow_overwrite = allow_overwrite
         self.props = props
+        self.to_json = append_metadata_to_json(  # type: ignore
+            self.to_json,
+            self.__class__,
+        )
+
+    def to_json(self) -> Dict:
```
lindong28, line 78: Do we need to add these two method for `FeathubJobDescriptor`? I suppose we don't need to store FeathubJobDescriptor in registry.

python/feathub/tests/feathub_it_test_base.py
```diff
@@ -47,6 +51,37 @@ def _merge_nested_dict(a, b) -> None:
...
+
+    def get_features(self, name: str) -> TableDescriptor:
+        return self.registry.get_features(name)
+
+    def delete_features(self, name: str) -> bool:
+        return self.delete_features(name)
+
+    @staticmethod
+    def _save_and_reload_through_json(features: TableDescriptor):
+        json_bytes = json.dumps(features.to_json(), sort_keys=True).encode("utf8")
```
lindong28, line 81: Can you explain why we need to use `encode("utf8")`? The Python doc does not seem to use this.

[1] https://docs.python.org/3/library/json.html

lindong28 (commented):

python/feathub/common/utils.py
```diff
@@ -85,3 +90,50 @@ def is_local_file_or_dir(url: str) -> bool:
...
+    :param __class__: The host class for the to_json method.
+    """
+
+    def wrapper(*args: Any, **kwargs: Any) -> Any:
+        json_dict = func(*args, **kwargs)
+        if "type" in json_dict or "version" in json_dict:
+            raise FeathubException(
+                f"f{__class__.__name__}#to_json should not contain metadata keys."
+            )
+        json_dict["type"] = __class__.__module__ + "." + __class__.__name__
```
lindong28, line 114: Can we rename the field to `class` to differentiate it from the `type` field used in the FeatHub config?

```
{
    "processor": {
        "type": "local",
    },
    "online_store": {
        "types": ["memory"],
        "memory": {},
    },
    "registry": {
        "type": "mysql",
        "mysql": {
            "database": "default",
            "table": "feathub_registry_features",
            "host": "127.0.0.1",
            "port": 3306,
            "username": "admin",
            "password": "123456",
        },
    },
    "feature_service": {
        "type": "local",
        "local": {},
    },
}
```

lindong28:
Thanks for the update! LGTM.

PR #139, opened 2023-05-22
[hotfix] Update RedisSink write and flush logic

yunfengzhou-hub:

In the current implementation, Redis sink uses pipeline to cache write requests which are flushed per checkpoint and when the cache is full. Such behavior has caused that in some use cases, some features might be deleted, and rewritten after dozens of seconds.

This pull request fixes this behavior by changing the behavior when the sink function writes out list and map data.

## Brief change log

- Compares lists and maps to be written out to Redis with existing values and only update the changed part.
- Flushes out data for every record.

  - The public API: no

## Documentation


yunfengzhou-hub:

PR #141, opened 2023-05-23
Introduce MAP and JSON_STRING builtin functions

Sxnan:

Introduce MAP and JSON_STRING builtin functions.

## Brief change log

- Introduce MAP and JSON_STRING builtin functions

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/feature_views/transforms/tests/test_expression_transform.py
+++ b/python/feathub/feature_views/transforms/tests/test_expression_transform.py
@@ -16,7 +16,7 @@ from abc import ABC
 import pandas as pd
 from dateutil.tz import tz
 
-from feathub.common.types import Float64, String, Int64
+from feathub.common.types import Float64, String, Int64, Bool, VectorType, MapType
 from feathub.common.utils import to_unix_timestamp
 from feathub.feature_views.derived_feature_view import DerivedFeatureView
 from feathub.feature_views.feature import Feature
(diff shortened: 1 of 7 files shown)


python/feathub/feature_views/transforms/tests/test_expression_transform.py
```diff
@@ -340,3 +340,113 @@ def test_duplicate_feature_name_with_source(self):
...
+                    name="row_map",
+                    transform="MAP('name', name, "
+                    "'cost', JSON_STRING(cost), "
+                    "'distance', JSON_STRING(distance), "
+                    "'time', time)",
+                ),
+                Feature(
+                    name="cost_distance_map",
+                    transform="MAP('cost', cost, 'distance', distance)",
+                ),
```
yunfengzhou-hub, line 362: Could you please also add a Feature whose `transform="MAP()"`? Zero arguments is also an even number of arguments.

yunfengzhou-hub (commented):

python/feathub/feature_views/transforms/tests/test_expression_transform.py
Sxnan: The problem with `MAP()` is that we cannot derive the type information of the result. Given that we don't see the use case at the moment, I think it is simpler that the `MAP` function requires at least two arguments.

Sxnan:
@yunfengzhou-hub Thanks for your review! I updated the PR accordingly. Can you take another look?

lindong28:
Thanks for the PR. LGTM.

PR #151, opened 2023-05-31
Run nightly with different python versions

Sxnan:

Run nightly with different python versions.

## Brief change log

- Run nightly with different python versions

  - The public API: no

## Documentation


Sxnan:
@yunfengzhou-hub Can you help review this PR?

.github/workflows/build-and-test.yaml
```diff
@@ -11,13 +11,23 @@ on:
         description: 'Run the build with tmate debugging enabled (https://github.com/marketplace/actions/debugging-with-tmate)'
         required: false
         default: false
+      python_version:
```
yunfengzhou-hub, line 14: Let's setup the default python version for `push` and `pull_request` as well.

.github/workflows/build-and-upload-wheel.yaml
```diff
@@ -30,12 +30,7 @@ jobs:
       - name: Set up python 3.7
         uses: actions/setup-python@v4
         with:
-          python-version: '3.7'
-      - name: Install Protoc
-        uses: arduino/setup-protoc@v1
-        with:
-          version: '3.17.3'
-          repo-token: ${{ secrets.FETCH_RELEASE_DATA }}
+          python-version: 3.7
```
yunfengzhou-hub, line 33: Shall we support different python versions here as well?

python/feathub/processors/flink/flink_processor.py
yunfengzhou-hub, line 1: Could you please verify whether Feathub tests can pass with pandas 2.0? It removed some API compared with 1.x, but given that the minimum supported Python version of pandas 2.0 is Python 3.8[1], this problem has not been considered in so far. I once encountered once such problem with Python 3.8 when developing in my local environment, but am not sure why Github Action did not reveal this.

[1] https://pandas.pydata.org/docs/getting_started/install.html#python-version-support

yunfengzhou-hub (commented):

.github/workflows/build-and-upload-wheel.yaml
Sxnan: Since we don't have any compiled binary, the built wheel is compatible with all Python3 versions. 

And we run examples against the built wheel in our nightly build with different Python versions.

python/feathub/processors/flink/flink_processor.py
Sxnan: I put an upper bound on pandas version "<2.0"

Sxnan:
@yunfengzhou-hub Thanks for your review! I updated the PR. Can you take another look?

yunfengzhou-hub:

python/setup.py
```diff
@@ -135,7 +135,7 @@ def remove_if_exists(file_path):
         package_data=PACKAGE_DATA,
         license="https://www.apache.org/licenses/LICENSE-2.0",
         author="FeatHub Authors",
-        python_requires=">=3.6",
+        python_requires=">=3.7",
```
lindong28, line 138: Can we also require that the version <= 3.9?

docs/examples/nyc_taxi.ipynb
```diff
@@ -51,7 +51,7 @@
    "source": [
     "## Install dependencies\n",
     "\n",
-    "This example has been verified in Python 3.7 with the following libraries.\n",
+    "This example has been verified in Python 3.7/3.8/3.9 with the following libraries.\n",
```
lindong28, line 54: Would it be simpler to avoid listing the supported python versions so that we don't need to update this sentence every time we update the supported python version?

lindong28 (commented):
Thanks for the PR. LGTM overall. Left two minor comments.

python/setup.py
Sxnan: We can. But I find that PyFlink doesn't specify the upper limit of the Python version so I think we should not put a tighter constraint on Feathub. And we can add the classifiers to specify the Python version, like [PyFlink](https://github.com/apache/flink/blob/cc552f970e2673f2895ca3c73a8150cdf720915a/flink-python/setup.py#L337). What do you think?

docs/examples/nyc_taxi.ipynb
Sxnan: I think we can avoid listing the supported Python version in the docs. But we cannot avoid listing them in the bash code, which also needs to be updated every time when we update the supported Python version, because Colab runs Python 3.10 by default and we need to install the compatible Python versions to run the notebook in Colab.

python/setup.py
lindong28: After thinking about this more, I agree it is better to skip specifying the version uppper-limit and follow the practice of pyflink/pyflink.

PR #154, opened 2023-06-08
Set HiveSource to streaming mode

yunfengzhou-hub:

This pull request fixes HiveSource's behavior and add test case and error message for certain unsupported cases.

## Brief change log

- Change HiveSource from batch to streaming read mode.
- Add error message and test case when HiveSource is created with timestamp.
- Add test case for the error message when a table without timestamp is used as the right table in a join

  - The public API: yes

## Documentation

### DIFF
--- a/python/feathub/feature_tables/sources/hive_source.py
+++ b/python/feathub/feature_tables/sources/hive_source.py
@@ -57,8 +57,6 @@ class HiveSource(FeatureTable):
         hive_catalog_conf_dir: str,
         schema: Schema,
         keys: Optional[List[str]] = None,
-        timestamp_field: Optional[str] = None,
-        timestamp_format: str = "epoch",
         processor_specific_props: Optional[Dict[str, str]] = None,
     ):
         """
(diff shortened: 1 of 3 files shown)


yunfengzhou-hub:
Hi @lindong28 could you please take a look at this PR?

python/feathub/feature_tables/sources/hive_source.py
```diff
@@ -75,9 +76,8 @@ def __init__(
         :param keys: Optional. The names of fields in this feature view that are
                      necessary to interpret a row of this table. If it is not None, it
                      must be a superset of keys of any feature in this table.
-        :param timestamp_field: Optional. If it is not None, it is the name of the field
-                                whose values show the time when the corresponding row
-                                is generated.
+        :param timestamp_field: Parameter inherited from parent class. Must be none for
```
lindong28, line 79: Would it be simpler to remove this field?

lindong28 (commented):
Thanks for the PR.

lindong28:
Thanks for the PR. LGTM.

PR #156, opened 2023-06-26
Optimize RedisSource map lookup behavior

yunfengzhou-hub:

This pull request mainly adds an optimization to Redis source so that when only certain keys of a map-typed feature would be used during a lookup join, the Redis source need not get the whole map value from Redis.

## Brief change log

- Support getItem operation in Feathub expression
- Support assigning default field names for features
- Support using getItem expression in JoinTransform
- Modify COLLECT_LIST's default value, changing from `[]` to `None` due to FLINK-32494.
- Optimize RedisSource in lookup joins under conditions described above.

  - The public API: no

## Documentation

### DIFF
--- a/java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/lookup/RedisLookupFunction.java
+++ b/java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/lookup/RedisLookupFunction.java
@@ -29,6 +29,8 @@ import org.apache.flink.table.types.DataType;
 import org.apache.flink.table.types.KeyValueDataType;
 import org.apache.flink.util.Preconditions;
 
+import org.apache.flink.shaded.jackson2.com.fasterxml.jackson.core.type.TypeReference;
+
 import com.alibaba.feathub.flink.connectors.redis.ConversionUtils;
 import com.alibaba.feathub.flink.connectors.redis.JedisClient;
 
@@ -37,9 +39,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import static com.alibaba.feathub.flink.connectors.redis.ConversionUtils.OBJECT_MAPPER;
+import static com.alibaba.feathub.flink.connectors.redis.RedisConfigs.HASH_FIELDS;
 import static com.alibaba.feathub.flink.connectors.redis.RedisConfigs.KEY_FIELDS;
 
 /**
@@ -75,6 +80,8 @@ public class RedisLookupFunction extends LookupFunction {
 
     private final DataType[] fieldDataTypes;
 
+    private final String[][] hashFields;
+
     private transient JedisClient client;
 
     private transient RowData.FieldGetter[] logicalKeyGetters;
@@ -87,6 +94,23 @@ public class RedisLookupFunction extends LookupFunction {
         this.featureFieldIndices = getFeatureFieldIndices(schema, config);
         this.physicalKeyFieldIndices = getPhysicalKeyFieldIndices(schema, featureFieldIndices);
         this.fieldDataTypes = schema.getColumnDataTypes().toArray(new DataType[0]);
+        if (config.get(HASH_FIELDS) == null) {
+            this.hashFields = null;
+        } else {
+            try {
+                TypeReference<HashMap<String, List<String>>> typeRef =
+                        new TypeReference<HashMap<String, List<String>>>() {};
+                HashMap<String, List<String>> map =
+                        OBJECT_MAPPER.readValue(config.get(HASH_FIELDS), typeRef);
+                this.hashFields = new String[schema.getColumnCount()][];
+                for (Map.Entry<String, List<String>> entry : map.entrySet()) {
+                    this.hashFields[schema.getColumnNames().indexOf(entry.getKey())] =
+                            entry.getValue().toArray(new String[0]);
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }
     }
 
     @Override
@@ -135,7 +159,16 @@ public class RedisLookupFunction extends LookupFunction {
                         ConversionUtils.fromList(redisData, (CollectionDataType) valueType);
                 result.setField(valueFieldIndex, flinkSqlData);
             } else if (valueType instanceof KeyValueDataType) {
-                Map<String, String> redisData = client.hgetAll(key);
+                Map<String, String> redisData;
+                if (hashFields == null || hashFields[valueFieldIndex] == null) {
+                    redisData = client.hgetAll(key);
+                } else {
+                    redisData = new HashMap<>();
+                    List<String> values = client.hmget(key, hashFields[valueFieldIndex]);
+                    for (int j = 0; j < hashFields[valueFieldIndex].length; j++) {
+                        redisData.put(hashFields[valueFieldIndex][j], values.get(j));
+                    }
+                }
                 MapData flinkSqlData =
                         ConversionUtils.fromMap(redisData, (KeyValueDataType) valueType);
                 result.setField(valueFieldIndex, flinkSqlData);
--- a/python/feathub/dsl/expr_utils.py
+++ b/python/feathub/dsl/expr_utils.py
@@ -11,14 +11,19 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
-from typing import Set
+
+import re
+from typing import Set, cast, Any, Tuple
 
 from ply import lex
 
+from feathub.dsl.ast import BracketOp, VariableNode, ValueNode
 from feathub.dsl.built_in_func import BUILTIN_FUNC_DEF_MAP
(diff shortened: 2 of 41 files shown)


python/feathub/feature_views/derived_feature_view.py
```diff
@@ -98,16 +102,34 @@ def build(
...
+                    force_update,
+                    f"f{default_field_name_index}",
                 )
+                if feature.name == f"f{default_field_name_index}":
+                    default_field_name_index += 1
             features.append(feature)
 
         self._validate(features, source)
 
+        self._resolve_join_feature_dtype(
```
lindong28, line 127: It seems a bit weird to do `self._resolve_join_feature_dtype` after `self._validate(...)`.

I suppose we need this `_resolve_join_feature_dtype` only if user specifies a string-valued expression. But if that is the case, should dtype already be resolved by the _get_feature_from_feature_str() (which is invoked above)?

lindong28 (commented):
Thanks for the PR!

Left some comments below.

python/feathub/dsl/expr_utils.py
```diff
@@ -34,3 +39,34 @@ def get_variables(feathub_expr: str) -> Set[str]:
...
+    """
+    ast = _parser.parse(expr)
+    return (
+        isinstance(ast, BracketOp)
+        and isinstance(ast.left_child, VariableNode)
+        and isinstance(ast.right_child, ValueNode)
+    )
+
+
+def get_map_static_lookup_key(expr: str) -> Any:
```
lindong28, line 66: How about renaming it as `get_static_map_lookup_key` for consistency with `is_static_map_lookup_op`?

python/feathub/feature_views/transforms/join_transform.py
```diff
@@ -25,26 +25,26 @@ class JoinTransform(Transformation):
     def __init__(
         self,
         table_name: str,
-        feature_name: str,
+        expr: str,
     ):
         """
         :param table_name: The name of a Source or FeatureView table.
-        :param feature_name: The feature name.
+        :param expr: The feature expr.
```
lindong28, line 32: Does this support arbitrary FeatHub expression? If no, can you add doc explaining what is supported?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -98,12 +104,24 @@ def build(
...
                 feature = self._get_feature_from_feature_str(
-                    feature, registry, source, force_update
+                    feature,
+                    registry,
+                    source,
+                    force_update,
+                    f"_{default_field_name_index}",
                 )
+                if feature.name == f"_{default_field_name_index}":
+                    default_field_name_index += 1
```
lindong28, line 124: It is possible that the this feature name collide with source.get_output_features().

Maybe we should put the logic above (which uses a while loop) into a method and call the method here?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -171,28 +190,44 @@ def _get_feature_from_feature_str(
...
-                "format {feature_name}."
-            )
+            elif is_static_map_lookup_op(parts[1]):
+                join_feature_expr = parts[1]
+                variable_types: Dict[str, Optional[DType]] = {
+                    f.name: f.dtype for f in table_desc.get_output_features()
+                }
+                return Feature(
+                    name=default_feature_name,
+                    dtype=cast(
```
lindong28, line 217: Since the logic of determing the dtype is non-trivial (involving 3+ function calls and spread across 3 lines), would it be more readable to determine it before constructing the Feature instance?

python/feathub/feature_views/feature_view.py
```diff
@@ -52,10 +52,11 @@ def __init__(
                        string, it should refer to the name of a table descriptor in the
                        registry.
         :param features: A list of features to be joined onto this feature view.
-                         If a feature is a string, it should be either in the format
-                         {table_name}.{feature_name}, which refers to a feature in the
-                         table with the given name, or in the format {feature_name},
-                         which refers to a feature in the source table.
+                         If a feature is a string, it should be it should be in either
+                         of the following formats:
+                         1. {feature_name}
```
lindong28, line 57: Can you add the comment to explain that `{feature_name}` refers to a feature in the source table?

Same for `OnDemandFeatureView`.

python/feathub/processors/flink/table_builder/flink_table_builder.py
```diff
@@ -408,6 +407,30 @@ def _get_table_from_derived_feature_view(
             right_table = table_by_names[right_table_name]
             right_table_descriptor = self.registry.get_features(right_table_name)
             redis_source = get_redis_source(right_table_descriptor)
+            right_table = optimize_redis_source_lookup_map(
```
lindong28, line 410: Would it be more readable to call this method only when `redis_source is not None`?

python/feathub/processors/flink/table_builder/redis_utils.py
```diff
@@ -96,9 +103,60 @@ def append_physical_key_columns_per_feature(
...
+    Redis hash, instead of reading all hash fields.
+    """
+    if not isinstance(table_descriptor, RedisSource):
+        return table
+
+    feature_names = [
+        x for x in table_descriptor.schema.field_names if x not in table_descriptor.keys
+    ]
+
+    if len(feature_names) != 1 or not isinstance(
```
lindong28, line 125: Is there particular reason that this optimization only support RedisSource with one field?

python/feathub/processors/flink/table_builder/flink_table_builder.py
```diff
@@ -408,6 +407,30 @@ def _get_table_from_derived_feature_view(
...
+                self.t_env,
+                right_table,
+                right_table_descriptor,
+                right_table_join_field_descriptors,
+            )
+
+            select_dict = {}
+            for field_name in right_table.get_schema().get_field_names():
+                select_dict[field_name] = f"`{field_name}`"
```
lindong28, line 419: Instead of enumerating all existing columns of the `right_table`, would it be simpler to only add the columns for those `right_table_join_field_descriptors` that contain "static map lookup op"? We can add columns similar to what is done in `append_physical_key_columns_per_feature()`.

And would it be more readable to move this logic into `join_utils#lookup_join`?

python/feathub/processors/flink/table_builder/redis_utils.py
```diff
@@ -96,9 +103,60 @@ def append_physical_key_columns_per_feature(
...
+        t_env,
+        table_descriptor,
+        {"hashFields": ",".join(x for x in hash_fields)},
+    )
+
+
 def get_table_from_redis_source(
     t_env: StreamTableEnvironment,
     source: RedisSource,
+    additional_options: Optional[Dict[str, str]] = None,
```
lindong28, line 159: How about using `extra_options` for consistency with `Feature`'s constructor?

python/feathub/feature_views/transforms/join_transform.py
```diff
@@ -25,26 +25,26 @@ class JoinTransform(Transformation):
     def __init__(
         self,
         table_name: str,
-        feature_name: str,
+        expr: str,
```
lindong28, line 28: Can you use `@deprecated_alias` to keep supporting `expr` for backward compatibility?

lindong28 (commented):
Thanks for the PR. Looks pretty good now. Left some comments below.

python/feathub/processors/flink/table_builder/flink_table_builder.py
yunfengzhou-hub: I have moved this logic to `redis_utils` instead of `join_utils` to avoid circular reference.

java/feathub-connectors/flink-connectors/flink-connector-redis/src/main/java/com/alibaba/feathub/flink/connectors/redis/lookup/RedisLookupFunction.java
```diff
@@ -87,6 +94,23 @@ public RedisLookupFunction(ReadableConfig config, ResolvedSchema schema) {
         this.featureFieldIndices = getFeatureFieldIndices(schema, config);
         this.physicalKeyFieldIndices = getPhysicalKeyFieldIndices(schema, featureFieldIndices);
         this.fieldDataTypes = schema.getColumnDataTypes().toArray(new DataType[0]);
+        if (config.get(HASH_FIELDS) == null || this.featureFieldIndices.length > 1) {
```
lindong28, line 97: Should we also support `HASH_FIELDS` when the Redis table only has more than one feature column?

python/feathub/processors/flink/table_builder/redis_utils.py
```diff
@@ -96,9 +107,125 @@ def append_physical_key_columns_per_feature(
...
 
 
+def optimize_redis_source_lookup_map(
+    t_env: StreamTableEnvironment,
+    table: NativeFlinkTable,
+    table_descriptor: TableDescriptor,
+    join_field_descriptors: Dict[str, JoinFieldDescriptor],
+) -> NativeFlinkTable:
+    """
+    If the right table of a lookup join is a RedisSource and only designated keys
```
lindong28, line 117: Does this mean that we do not support filtering-by-map-key when there are columns with non-map type?

lindong28 (commented):

python/feathub/processors/flink/table_builder/redis_utils.py
yunfengzhou-hub: Filtering by map key is still supported when there are non-map-typed columns. There might be some ambiguity in this description. I'll improve this document.

lindong28:
Thanks for the update. LGTM.

PR #157, opened 2023-06-27
Support keep_timestamp_field in Sinks

yunfengzhou-hub:

This pull request adds `keep_timestamp_field` in Sinks' constructors. The default value of this parameter is True, and the corresponding behavior matches the existing implementation. When the parameter is set to False, timestamp field would be removed before persisting to external systems using the Sinks.

## Brief change log

- Add `keep_timestamp_field` in Sinks' constructor
- Modify the signature of methods related to materialization, changing the data type of the sink parameter to Sink class.
- Fix the problem that processing time attribute is left unremoved after Redis lookup join.

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? python docs

### DIFF
--- a/python/feathub/feature_tables/tests/test_redis_source_sink.py
+++ b/python/feathub/feature_tables/tests/test_redis_source_sink.py
@@ -32,6 +32,7 @@ from feathub.feature_tables.sources.redis_source import (
 from feathub.feature_views.derived_feature_view import DerivedFeatureView
 from feathub.feature_views.feature import Feature
 from feathub.feature_views.transforms.join_transform import JoinTransform
+from feathub.feature_views.transforms.python_udf_transform import PythonUdfTransform
 from feathub.online_stores.conversion_utils import to_python_object
 from feathub.table.schema import Schema
 from feathub.tests.feathub_it_test_base import FeathubITTestBase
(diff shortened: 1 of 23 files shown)


yunfengzhou-hub:

python/feathub/feature_tables/tests/test_redis_source_sink.py
```diff
@@ -404,6 +404,65 @@ def _test_redis_sink_custom_key(
...
 
 
+def _test_redis_sink_not_keep_timestamp(
+    self: FeathubITTestBase,
+    host: str,
+    port: int,
+    mode: str,
+    redis_client: Union[Redis, RedisCluster],
+):
+    # TODO: Fix the bug that in flink processor when column "val"
```
lindong28, line 414: Can you create a JIRA and provide JIRA in this comment?

lindong28 (commented):
Thanks for the PR. LGTM overall. Left one minor comment.

python/feathub/feature_tables/tests/test_redis_source_sink.py
yunfengzhou-hub: It seems that this problem no longer exist in Feathub after previous code updates and Flink version upgrade. I have removed all such TODOs and add test case that verify output results when the input contains None values.

lindong28:
Thanks for the update. LGTM.

PR #159, opened 2023-07-03
Fix join expr for single field names

yunfengzhou-hub:

This pull request fixes the bug that exception would be thrown when field names that equal to reserved keywords are used as keys in JoinTransforms.

## Brief change log

- Fixes the expression for join field descriptors to support cases when the field names are equal to reserved keywords.
- Fixes the bug that `expr_utils.get_var_name` cannot correct extract field name from expressions like "\`field_name\`".

  - The public API: no

## Documentation


lindong28:
Thanks for the PR! LGTM.

PR #160, opened 2023-07-04
Skip empty map when enable partal update in RedisSink

yunfengzhou-hub:

This pull request skips deleting Redis hash entry when `enable_hash_partial_update` is True and the input is None or an empty map.

## Brief change log

- Avoids deleting Redis hash entry when `enable_hash_partial_update` is True and the input is None or an empty map.

  - The public API: yes

## Documentation


lindong28:
Thanks for the PR. LGTM.

PR #162, opened 2023-07-13
Introduce Flink Prometheus sink

Sxnan:

Introduce Flink Prometheus sink

## Brief change log

- Introduce Flink Prometheus sink

  - The public API: no

## Documentation

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -61,7 +61,7 @@ jobs:
         run: |
           # notebooks should not contain outputs
           jupyter nbconvert --clear-output --inplace docs/examples/*.ipynb
-          if [[ `git status --porcelain` ]]; then
+          if [[ `git status --porcelain docs/examples` ]]; then
             exit 1
           fi
 
--- a/java/feathub-connectors/flink-connectors/flink-connector-prometheus/pom.xml
+++ b/java/feathub-connectors/flink-connectors/flink-connector-prometheus/pom.xml
@@ -1,0 +1,125 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2022 The FeatHub Authors
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~     https://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>com.alibaba.feathub</groupId>
+        <artifactId>flink-connectors</artifactId>
+        <version>0.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>flink-connector-prometheus</artifactId>
+
+    <properties>
+        <prometheus.client.version>0.8.1</prometheus.client.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-api-java-bridge</artifactId>
+            <version>${flink.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.prometheus</groupId>
+            <artifactId>simpleclient_pushgateway</artifactId>
+            <version>${prometheus.client.version}</version>
+        </dependency>
+
+        <!-- test dependencies -->
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers</artifactId>
+            <version>1.16.2</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-connector-test-utils</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-planner-loader</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-table-runtime</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-clients</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
+            <artifactId>flink-csv</artifactId>
+            <version>${flink.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-slf4j-impl</artifactId>
+            <version>2.19.0</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
+                <version>3.4.1</version>
+                <executions>
+                    <execution>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>shade</goal>
+                        </goals>
+                        <configuration>
+                            <artifactSet>
+                                <includes>
+                                    <include>io.prometheus:*</include>
+                                    <include>javax.xml.bind:jaxb-api</include>
+                                </includes>
+                            </artifactSet>
+                            <relocations>
+                                <relocation>
+                                    <pattern>io.prometheus</pattern>
+                                    <shadedPattern>
+                                        com.alibaba.feathub.connectors.flink.shaded.io.prometheus
+                                    </shadedPattern>
+                                </relocation>
+                            </relocations>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+</project>
--- a/java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/test/java/com/alibaba/feathub/flink/connectors/prometheus/PrometheusSinkITTest.java
+++ b/java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/test/java/com/alibaba/feathub/flink/connectors/prometheus/PrometheusSinkITTest.java
@@ -1,0 +1,509 @@
+/*
+ * Copyright 2022 The FeatHub Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.alibaba.feathub.flink.connectors.prometheus;
+
+import org.apache.flink.streaming.api.environment.StreamExecutionEnvironment;
+import org.apache.flink.table.api.DataTypes;
+import org.apache.flink.table.api.Schema;
+import org.apache.flink.table.api.Table;
+import org.apache.flink.table.api.TableDescriptor;
+import org.apache.flink.table.api.bridge.java.StreamTableEnvironment;
+import org.apache.flink.types.Row;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ExecutionException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/** Tests Prometheus sink. */
+public class PrometheusSinkITTest extends PrometheusTableTestBase {
+
+    private StreamExecutionEnvironment env;
+    private StreamTableEnvironment tEnv;
+
+    @BeforeEach
+    void setUp() {
+        env = StreamExecutionEnvironment.getExecutionEnvironment();
+        tEnv = StreamTableEnvironment.create(env);
+    }
+
+    @Test
+    void testSink() throws ExecutionException, InterruptedException, IOException {
+        final Table sourceTable =
+                tEnv.fromDataStream(env.fromElements(Row.of(0), Row.of(99))).as("metric_val");
+
+        tEnv.createTemporaryView("sourceTable", sourceTable);
+
+        final Table metricTable =
+                tEnv.sqlQuery(
+                        "SELECT CAST(metric_val AS TINYINT) AS tinyint_val, "
+                                + "CAST(metric_val AS SMALLINT) AS smallint_val, "
+                                + "CAST(metric_val AS INTEGER) AS int_val, "
+                                + "CAST(metric_val AS BIGINT) AS bigint_val, "
+                                + "CAST(metric_val AS DECIMAL) AS decimal_val, "
+                                + "CAST(metric_val AS FLOAT) AS float_val, "
+                                + "CAST(metric_val AS DOUBLE) AS double_val FROM sourceTable");
+
+        final String serverUrl = PROMETHEUS_PUSH_GATEWAY_CONTAINER.getHostUrl();
+        metricTable
+                .executeInsert(
+                        TableDescriptor.forConnector("prometheus")
+                                .option("serverUrl", serverUrl)
+                                .option("deleteOnShutdown", "false")
+                                .option("jobName", "testSink")
+                                .option("extraLabels", "k1=v1,k2=v2")
+                                .build())
+                .await();
+
+        final HashMap<String, List<Metric>> metrics = getMetrics(serverUrl);
+        final HashMap<String, String> expectedLabel = new HashMap<>();
+        expectedLabel.put("job", "\"testSink\"");
+        expectedLabel.put("instance", "\"\"");
+        expectedLabel.put("k1", "\"v1\"");
+        expectedLabel.put("k2", "\"v2\"");
+
+        checkMetric(metrics.get("tinyint_val").get(0), "tinyint_val", "99", expectedLabel);
+        checkMetric(metrics.get("smallint_val").get(0), "smallint_val", "99", expectedLabel);
+        checkMetric(metrics.get("int_val").get(0), "int_val", "99", expectedLabel);
+        checkMetric(metrics.get("bigint_val").get(0), "bigint_val", "99", expectedLabel);
+        checkMetric(metrics.get("decimal_val").get(0), "decimal_val", "99", expectedLabel);
+        checkMetric(metrics.get("float_val").get(0), "float_val", "99", expectedLabel);
+        checkMetric(metrics.get("double_val").get(0), "double_val", "99", expectedLabel);
+    }
+
+    @Test
+    void testSinkColumnWithComment() throws ExecutionException, InterruptedException, IOException {
+        final Table sourceTable =
+                tEnv.fromDataStream(env.fromElements(Row.of(0), Row.of(99))).as("metric_val");
+
+        tEnv.createTemporaryView("sourceTable", sourceTable);
+
+        final Table metricTable =
+                tEnv.sqlQuery(
+                        "SELECT CAST(metric_val AS TINYINT) AS tinyint_val, "
+                                + "CAST(metric_val AS SMALLINT) AS smallint_val, "
+                                + "CAST(metric_val AS INTEGER) AS int_val, "
+                                + "CAST(metric_val AS BIGINT) AS bigint_val, "
+                                + "CAST(metric_val AS DECIMAL) AS decimal_val, "
+                                + "CAST(metric_val AS FLOAT) AS float_val, "
+                                + "CAST(metric_val AS DOUBLE) AS double_val FROM sourceTable");
+
+        final String serverUrl = PROMETHEUS_PUSH_GATEWAY_CONTAINER.getHostUrl();
+        metricTable
+                .executeInsert(
+                        TableDescriptor.forConnector("prometheus")
+                                .schema(
+                                        Schema.newBuilder()
+                                                .column("tinyint_val", DataTypes.TINYINT())
+                                                .withComment(
+                                                        "value_type=tinyint,col_k=col_v,test_label=\\;")
+                                                .column("smallint_val", DataTypes.SMALLINT())
+                                                .withComment("value_type=smallint,col_k=col_v")
+                                                .column("int_val", DataTypes.INT())
+                                                .withComment("value_type=int,col_k=col_v")
+                                                .column("bigint_val", DataTypes.BIGINT())
+                                                .withComment("value_type=bigint,col_k=col_v")
+                                                .column("decimal_val", DataTypes.DECIMAL(10, 2))
+                                                .withComment("value_type=decimal,col_k=col_v")
+                                                .column("float_val", DataTypes.FLOAT())
+                                                .withComment("value_type=float,col_k=col_v")
+                                                .column("double_val", DataTypes.DOUBLE())
+                                                .withComment("value_type=double,col_k=col_v")
+                                                .build())
+                                .option("serverUrl", serverUrl)
+                                .option("deleteOnShutdown", "false")
+                                .option("jobName", "testSink")
+                                .option("extraLabels", "k1=v1,k2=v2")
+                                .build())
+                .await();
+
+        final HashMap<String, List<Metric>> metrics = getMetrics(serverUrl);
+        final HashMap<String, String> tableLabels = new HashMap<>();
+        tableLabels.put("job", "\"testSink\"");
+        tableLabels.put("instance", "\"\"");
+        tableLabels.put("k1", "\"v1\"");
+        tableLabels.put("k2", "\"v2\"");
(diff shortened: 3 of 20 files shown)


Sxnan:
@yunfengzhou-hub Can you review this PR?

yunfengzhou-hub (commented):

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,181 @@
...
+                case BIGINT:
+                    metric.set(row.getLong(i));
+                    break;
+                case FLOAT:
+                    metric.set(row.getFloat(i));
+                    break;
+                case DOUBLE:
+                    metric.set(row.getDouble(i));
+                    break;
+                default:
```
yunfengzhou-hub, line 135: In some scenarios, a feature might only have a limited number of possible values, and users want to count the ratio for each label. 

In order to support this scenario, let's also add support for `Map<String, Number>` types. Each entry corresponds to a metric in Prometheus, the key in an entry is a label of the metric, and the value of the entry is the value of the metric.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,181 @@
...
+        metrics = new Gauge[columnNames.size()];
+        for (int i = 0; i < columnNames.size(); i++) {
+            final LogicalType logicalType = fieldTypes.get(i).getLogicalType();
+            if (!logicalType.is(LogicalTypeFamily.NUMERIC)) {
+                throw new IOException("PrometheusSink only support numeric data types.");
+            }
+
+            metrics[i] =
+                    Gauge.build()
+                            .name(columnNames.get(i))
```
yunfengzhou-hub, line 99: It might be better to decouple column names from prometheus metric names, so that we can write to Prometheus metrics with the same name and different labels.

yunfengzhou-hub (commented):
Add more comments according to further offline discussion.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,57 @@
...
+            ConfigOptions.key("deleteOnShutdown")
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
+                            Description.builder()
+                                    .text(
+                                            "Specifies whether to delete metrics from the PushGateway on shutdown."
+                                                    + " Flink will try its best to delete the metrics but this is not guaranteed.")
+                                    .build());
+    public static final ConfigOption<String> EXTRA_LABELS =
```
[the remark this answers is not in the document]
Sxnan: Thanks for the suggestion. I updated the PR so that users can configure metadata, including labels, of a column with column comment.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,181 @@
...
+                case DECIMAL:
+                    final int precision = ((DecimalType) type).getPrecision();
+                    final int scale = ((DecimalType) type).getScale();
+                    final DecimalData decimal = row.getDecimal(i, precision, scale);
+                    metric.set(decimal.toBigDecimal().doubleValue());
+                    break;
+                case TINYINT:
+                    metric.set(row.getByte(i));
+                    break;
+                case SMALLINT:
```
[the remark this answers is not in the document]
Sxnan: I think supporting the additional type of decimal, tinyint and smallint doesn't add much complexity to the code base and it can be more future-proof, i.e. Feathub may support those types in the future. What do you think?

Sxnan:
@yunfengzhou-hub Thanks for the review! I updated the PR accordingly. Could you take another look?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,353 @@
...
+        super.open(parameters);
+
+        registry = new CollectorRegistry();
+        pushGateway = new PushGateway(config.get(HOST_URL));
+        jobName = config.get(JOB_NAME);
+        extraLabels = parseGroupingKey(config.get(EXTRA_LABELS));
+
+        for (int i = 0; i < metricMetas.size(); i++) {
+            final LogicalType logicalType = fieldTypes.get(i).getLogicalType();
+            verifyType(logicalType);
```
yunfengzhou-hub, line 121: Let's also verify that metrics with the same name should have the same label names. It seems that user would get an exception that says "the length of label names and labels mismatch" when the above rule is violated in the current implementation, which is not straight forward enough to know how to correct the mistake.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    static Map<String, String> parseGroupingKey(final String groupingKeyConfig) {
+        if (!groupingKeyConfig.isEmpty()) {
+            Map<String, String> groupingKey = new HashMap<>();
+            String[] kvs = groupingKeyConfig.split(";");
+            for (String kv : kvs) {
+                int idx = kv.indexOf("=");
+                if (idx < 0) {
+                    LOG.warn("Invalid prometheusPushGateway groupingKey:{}, will be ignored", kv);
```
yunfengzhou-hub, line 257: It might be better to throw exception instead of log a warning when this case happens.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+                            </relocations>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
+
+
+
```
yunfengzhou-hub, line 125: nit: too much blank lines here.

.github/workflows/build-and-test.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: |
           # notebooks should not contain outputs
           jupyter nbconvert --clear-output --inplace docs/examples/*.ipynb
-          if [[ `git status --porcelain` ]]; then
+          if [[ `git status --porcelain docs/examples` ]]; then
```
yunfengzhou-hub, line 64: I know it is reasonable to make this change, but could you please briefly illustrate how the previous implementation has caused failures in Github workflows?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
yunfengzhou-hub, line 49: nit: It might be better to add a blank line above.

yunfengzhou-hub (commented):
Thanks for the update. Left some comments as below.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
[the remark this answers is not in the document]
yunfengzhou-hub: I'm OK with it given that we have added test cases that cover these data types in this PR.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,353 @@
...
+            final MetricMeta<Gauge.Child> metricMeta = metricMetas.get(i);
+            final String metricName = metricMeta.name;
+
+            final SimpleCollector<Gauge.Child> collector =
+                    nameToCollectors.computeIfAbsent(
+                            metricName,
+                            key ->
+                                    Gauge.build()
+                                            .name(key)
+                                            .help(key)
```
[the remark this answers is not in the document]
Sxnan: Unfortunately, `help` is required by `Prometheus`.

.github/workflows/build-and-test.yaml
Sxnan: The Prometheus connector module uses a shade plugin to build the fat jar, producing a `dependency-reduced-pom.xml`. The `git status --porcelain` will complain about the new file `dependency-reduced-pom.xml`.

Sxnan:
@yunfengzhou-hub Thanks for the review! I updated the PR. Could you take another look?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,58 @@
...
+                                    .build());
+
+    public static final ConfigOption<String> EXTRA_LABELS =
+            ConfigOptions.key("extraLabels")
+                    .stringType()
+                    .defaultValue("")
+                    .withDescription(
+                            "The extra labels of all the metrics. The label name and value are separated"
+                                    + " by '=', and labels are separated by ';', e.g., k1=v1;k2=v2. Please "
+                                    + "ensure that your grouping key meets the Prometheus requirements.");
```
yunfengzhou-hub, line 57: nit: Let's use "grouping key" or "extra label" uniformly across all documents and variable names.

yunfengzhou-hub (commented):
Thanks for the update. LGTM overall. Left one small comment as below.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,59 @@
...
+                                            "Specifies whether to delete metrics from the PushGateway on shutdown."
+                                                    + " Flink will try its best to delete the metrics but this is not guaranteed.")
+                                    .build());
+
+    public static final ConfigOption<String> GROUPING_KEY =
+            ConfigOptions.key("groupingKey")
+                    .stringType()
+                    .defaultValue("")
+                    .withDescription(
+                            "Specifies the grouping key which is the group and global labels of all metrics pushed by this sink. "
```
lindong28, line 55: What is the purpose/meaning of mentioning `global` here?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,59 @@
...
+                    .booleanType()
+                    .defaultValue(true)
+                    .withDescription(
+                            Description.builder()
+                                    .text(
+                                            "Specifies whether to delete metrics from the PushGateway on shutdown."
+                                                    + " Flink will try its best to delete the metrics but this is not guaranteed.")
+                                    .build());
+
+    public static final ConfigOption<String> GROUPING_KEY =
```
lindong28, line 50: It is mentioned in the PushGateway README [1] that "Pushed metrics are managed in groups, identified by a grouping key of any number of labels, of which the first must be the job label".

Therefore, the grouping key used in the PushGateway is expected to have "job" as the first key. In contrast, the config specified here will not have "job" as the first key. It might be confusing to use the same terminology with different specifications.

How about renaming it `extraLabels`?

[1] https://github.com/prometheus/pushgateway

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,366 @@
...
+import static com.alibaba.feathub.flink.connectors.PrometheusConfigs.JOB_NAME;
+
+/**
+ * A {@link org.apache.flink.streaming.api.functions.sink.SinkFunction} that writes to a Prometheus
+ * PushGateway. Note that the SinkFunction only support writing numeric columns or map columns that
+ * map from string to numeric values. Each column can be configured with a json map via the column
+ * comment. Below are the configuration of the column:
+ *
+ * <ul>
+ *   <li>metric_name: This specifies the metric name under which the column value is reported. If
```
lindong28, line 68: I am wondering which use-case requires this capability of renaming the metric. Can you describe such a use-case?

It is mentioned in the PushGateway README that "Metrics of the same name must have the same type". So we probably have to encode some of the "tags" in the metric name so that we can derive metrics with different type from the same feature anyway.

lindong28 (commented):

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/test/java/com/alibaba/feathub/flink/connectors/prometheus/PrometheusSinkITTest.java
```diff
@@ -0,0 +1,454 @@
...
+            res.put(keyValues[i], keyValues[i + 1]);
+        }
+        return res;
+    }
+
+    private void checkMetric(
+            Metric metric, String name, String value, Map<String, String> expectedLabels) {
+        assertThat(metric.getName()).isEqualTo(name);
+        assertThat(metric.getValue()).isEqualTo(value);
+        assertThat(metric.getLabels()).containsAllEntriesOf(expectedLabels);
```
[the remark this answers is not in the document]
Sxnan: According to README[1], the pushgateway will attach a `job` label and an `instance` label to each scraped metric. I think we only need to ensure the labels sent by the sink are expected.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/test/java/com/alibaba/feathub/flink/connectors/prometheus/PrometheusSinkITTest.java
```diff
@@ -0,0 +1,454 @@
...
+
+        final String hostUrl = PROMETHEUS_PUSH_GATEWAY_CONTAINER.getHostUrl();
+        metricTable
+                .executeInsert(
+                        TableDescriptor.forConnector("prometheus")
+                                .schema(
+                                        Schema.newBuilder()
+                                                .column("tinyint_val", DataTypes.TINYINT())
+                                                .withComment(
+                                                        "{\"metric_name\":\"tinyint_metric\",\"value_type\":\"tinyint\"}")
```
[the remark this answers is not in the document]
Sxnan: According to the json spec[1],  string are delimited with double quotation marks and support a backslash [escaping](https://en.wikipedia.org/wiki/Escape_character) syntax.

If we want to make it simpler, we have to parse the configuration ourself, which is what I try to avoid.

[1] https://en.wikipedia.org/wiki/JSON

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
Sxnan: This is required by the Feathub as Feathub metrics all have the same name, such as count, ratio, etc. They are differentiated only by the label value. And Flink table cannot have two columns with the same name. Therefore, we need to have a way to map different columns to the same metric name.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,59 @@
...
+                            "The PushGateway server host URL including scheme, host name, and port.");
+
+    public static final ConfigOption<String> JOB_NAME =
+            ConfigOptions.key("jobName")
+                    .stringType()
+                    .noDefaultValue()
+                    .withDescription("The job name under which metrics will be pushed.");
+
+    public static final ConfigOption<Boolean> DELETE_ON_SHUTDOWN =
+            ConfigOptions.key("deleteOnShutdown")
```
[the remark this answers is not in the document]
Sxnan: According to this issue [1]

Sxnan:
@lindong28 I updated the PR according to our offline discussion. Could you take another look?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    @Override
+    public void open(Configuration parameters) throws Exception {
+        super.open(parameters);
+        registry = new CollectorRegistry();
+        pushGateway = new PushGateway(config.get(SERVER_URL));
+
+        for (final MetricMeta metricMeta : metricMetas) {
+            final String metricName = metricMeta.name;
```
lindong28, line 132: Can we use either `name` or `metricName` across this PR for consistency?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+        this.config = config;
+        this.fieldTypes = schema.getColumnDataTypes();
+        this.nameToCollectors = new HashMap<>();
+        this.jobName = config.get(JOB_NAME);
+        this.extraLabels = parseLabels(config.get(EXTRA_LABELS));
+
+        final List<Column> columns = schema.getColumns();
+        this.metricMetas = new ArrayList<>(columns.size());
+
+        Map<String, Set<String>> metricNameToLabelNames = new HashMap<>();
```
lindong28, line 107: Can we make this variable's name consistent with `nameToCollectors`?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+        super.open(parameters);
+        registry = new CollectorRegistry();
+        pushGateway = new PushGateway(config.get(SERVER_URL));
+
+        for (final MetricMeta metricMeta : metricMetas) {
+            final String metricName = metricMeta.name;
+            final SimpleCollector<Gauge.Child> collector =
+                    nameToCollectors.computeIfAbsent(
+                            metricName,
+                            key ->
```
lindong28, line 136: Here we use `name` (as prefix of `nameToCollectors`), `metricName` and `key` for the same thing, and then uses it as parameter of `..name(key)` below. This makes code inconsistent and hard to read.

Can you go over this PR and fix inconsistencies like this?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+        void setCollector(SimpleCollector<Gauge.Child> collector) {
+            this.collector = collector;
+        }
+
+        Gauge.Child getChild(String keyLabel) {
+            return children.computeIfAbsent(
+                    keyLabel,
+                    label -> {
+                        final ArrayList<String> allLabels = new ArrayList<>(this.labels);
+                        for (int i = 0; i < keyLabelCnt; ++i) {
```
lindong28, line 333: Will `keyLabelCnt` always be `1` for map-typed value? If so, would it be better to make this requirement explicit in the code and in the document?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+                    nullLabelNames);
+        }
+
+        void setCollector(SimpleCollector<Gauge.Child> collector) {
+            this.collector = collector;
+        }
+
+        Gauge.Child getChild(String keyLabel) {
+            return children.computeIfAbsent(
+                    keyLabel,
```
lindong28, line 330: It seems not easy to understand what is "key label".

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+                labelNames.put(labelName, labelValue);
+            }
+
+            return labelNames;
+        }
+
+        return Collections.emptyMap();
+    }
+
+    private static class MetricMeta implements Serializable {
```
lindong28, line 282: It seems `metadata` is very rarely abbreviated as `meta` (if any) in Flink or other project. Would it be better to use `MetricMetadata` here?

lindong28 (commented):

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusConfigs.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.apache.flink.configuration.ConfigOption;
+import org.apache.flink.configuration.ConfigOptions;
+import org.apache.flink.configuration.description.Description;
+
+/** Configurations used by Prometheus sink. */
+public class PrometheusConfigs {
+
+    public static final ConfigOption<String> HOST_URL =
+            ConfigOptions.key("hostUrl")
```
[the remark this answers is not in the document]
lindong28: It appears that this config name is not updated.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,351 @@
...
+import static com.alibaba.feathub.flink.connectors.PrometheusConfigs.DELETE_ON_SHUTDOWN;
+import static com.alibaba.feathub.flink.connectors.PrometheusConfigs.EXTRA_LABELS;
+import static com.alibaba.feathub.flink.connectors.PrometheusConfigs.JOB_NAME;
+import static com.alibaba.feathub.flink.connectors.PrometheusConfigs.SERVER_URL;
+
+/**
+ * A {@link org.apache.flink.streaming.api.functions.sink.SinkFunction} that writes to a Prometheus
+ * PushGateway. Note that the SinkFunction only support writing numeric columns or map columns that
+ * map from string to numeric values. User can specify extra labels of the column via the column
+ * comment. The label name and value are separated by '=', and labels are separated by ';', e.g.,
```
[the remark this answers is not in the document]
Sxnan: I just follow the config `extraLabels` and the config `groupingKey` of the Flink Prometheus Metric Reporter[1]. 

As per our offline discussion, we will support list column type and a list of pairs of kv. We can make it in the following format [k1=v11;k2=v21,k1=v12;k2=v22]. What do you think?

[1] https://nightlies.apache.org/flink/flink-docs-release-1.17/docs/deployment/metric_reporters/#groupingkey

Sxnan:
@lindong28 Thanks for your review! I updated the PR accordingly. Could you take another look?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,342 @@
...
+ * <p>The behavior of the Prometheus Sink for different column types is as follows:
+ *
+ * <ul>
+ *   <li>Numeric Column: Each column is emitted as a separate metric. Users can specify zero or one
+ *       label set for the metric to be emitted.
+ *   <li>List Column: Each element in the list is emitted as a separate metric. Users must specify
+ *       the same number of label sets as the number of elements in the list. Each label set
+ *       represents the corresponding element's labels. Additionally, all label sets must have the
+ *       same set of keys.
+ *   <li>Map Column: Each entry in the map is emitted as a separate metric. Users must specify
```
lindong28, line 70: How about using the following document:

The value type of this column must be a map<string, double>. For each key in this map, a metric is emitted with metric name equal to column's name, metric values equal to the values of the given key. The metric labels are derived from the column's comment as specified above, except that any label whose value is "null" in the column's comment should be updated to use the given key as the label value.

Note that it is not necessarily to explain the requirement of "having at least one label with null value" because 1) we will throw exception in this case with clear message and 2) for cases where user wants to use Map Column, they will always need to use null, and therefore they should not be surprised if we throw exception without documenting this. It is not necessary (and impossible) to document all possible exception in the Java doc.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,342 @@
...
+
+    @Override
+    public void invoke(RowData row, Context context) throws Exception {
+        for (int i = 0; i < columnInfos.size(); i++) {
+            setMetric(row, i);
+        }
+        pushGateway.pushAdd(registry, jobName, extraLabels);
+    }
+
+    public void setMetric(RowData row, int index) throws IOException {
```
lindong28, line 141: Would it be better to rename this method `updateMetric` for consistency with e.g. `MetricUpdater`?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/PrometheusSinkFunction.java
```diff
@@ -0,0 +1,342 @@
...
+    @Override
+    public void close() throws Exception {
+        super.close();
+        if (deleteOnShutdown) {
+            pushGateway.delete(jobName);
+        }
+    }
+
+    private static Map<String, String> parseLabels(final String labelsConfig) {
+        if (!labelsConfig.isEmpty()) {
```
lindong28, line 239: Would it be simpler to remove this `if` statement?

lindong28 (commented):

Sxnan:
@lindong28 Thanks for your review! I updated the PR accordingly. Could you take another look?

lindong28:
Thanks for the update. LGTM!

PR #169, opened 2023-07-19
Introduce MetricStore framework and add PrometheusMetricStore

yunfengzhou-hub:

This pull request adds MetricStore framework and PrometheusMetricStore.

## Brief change log

- Add MetricStore and Metric API
- Add PrometheusMetricStore

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? python docs

### DIFF
--- a/python/feathub/metric_stores/metric.py
+++ b/python/feathub/metric_stores/metric.py
@@ -1,0 +1,197 @@
+# Copyright 2022 The FeatHub Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import collections
+from abc import ABC, abstractmethod
+from datetime import timedelta
+from typing import Optional, Dict, OrderedDict, Sequence, Callable
+
+from feathub.common.exceptions import FeathubException
+from feathub.common.utils import append_metadata_to_json
+from feathub.feature_views.transforms.expression_transform import ExpressionTransform
+from feathub.feature_views.transforms.sliding_window_transform import (
+    SlidingWindowTransform,
+)
+from feathub.feature_views.transforms.transformation import Transformation
+
+
+# TODO: Add document to record built-in metrics.
+class Metric(ABC):
+    """
+    Common interface for all metrics. A metric refers to a statistic of a
+    characteristic of a feature.
+    """
+
+    def __init__(self, metric_type: str, window_size: timedelta = timedelta(seconds=0)):
+        """
+        :param metric_type: The type of the metric.
+        :param window_size: The time range to compute the metric. It should be zero or
+                            a positive time span. If it is zero, the metric will be
+                            computed from all features that have been processed since
+                            the Feathub job is created.
+        """
+        if window_size <= timedelta(seconds=0):
+            raise FeathubException(
+                f"Metric window size {window_size} must be a positive value."
+            )
+
+        self.metric_type = metric_type
+        self.window_size = window_size
+
+    def get_tags(self) -> OrderedDict:
+        """
+        Gets the tags of this metric for tag-based metric reporters.
+        """
+        return collections.OrderedDict(
+            [("window_time_sec", str(int(self.window_size / timedelta(seconds=1))))]
+        )
+
+    @abstractmethod
+    def get_transform_functions(self) -> Sequence[Callable[[str], Transformation]]:
+        """
+        Gets the functions to generate transformations used to aggregate and compute the
+        metric value. The function has the following parameter:
+
+        - feature_name: Name of a variable containing the intermediate metric value
+                        generated from previously defined transformation. For the first
+                        transformation, its value is the host feature name.
+        """
+        pass
+
+    @abstractmethod
+    def to_json(self) -> Dict:
+        """
+        Returns a json-formatted object representing this metric.
+        """
+        pass
+
+
+class Count(Metric):
+    """
+    A Metric that shows the number of features.
+    """
+
+    def __init__(
+        self,
+        filter_expr: Optional[str] = None,
+        window_size: timedelta = timedelta(seconds=0),
+    ):
+        """
+        :param filter_expr: Optional. If it is not None, it represents a partial FeatHub
+                            expression which evaluates to a boolean value. The partial
+                            Feathub expression should be a binary operator whose left
+                            child is absent and would be filled in with the host feature
+                            name. For example, "IS NULL" will be enriched into
+                            "{feature_name} IS NULL". Only features that evaluate this
+                            expression into True will be considered when computing the
+                            metric.
+        :param window_size: The time range to compute the metric. It should be zero or
+                            a positive time span. If it is zero, the metric will be
+                            computed from all features that have been processed since
+                            the Feathub job is created.
+        """
+        super().__init__("count", window_size)
+        self.filter_expr = filter_expr
+
+    def get_tags(self) -> OrderedDict:
+        tags = super(Count, self).get_tags()
+        tags["filter_expr"] = "" if self.filter_expr is None else self.filter_expr
+        return tags
+
+    def get_transform_functions(self) -> Sequence[Callable[[str], Transformation]]:
+        return [
+            lambda feature_name: SlidingWindowTransform(
+                expr="1",
+                agg_func="COUNT",
+                window_size=self.window_size,
+                step_size=self.window_size,
+                filter_expr=None
+                if self.filter_expr is None
+                else f"{feature_name} {self.filter_expr}",
+            )
+        ]
+
+    @append_metadata_to_json
+    def to_json(self) -> Dict:
+        return {
+            "window_size_sec": self.window_size / timedelta(seconds=1),
+            "filter_expr": self.filter_expr,
+        }
+
+    @classmethod
+    def from_json(cls, json_dict: Dict) -> "Count":
+        return Count(
+            filter_expr=json_dict["filter_expr"],
+            window_size=timedelta(seconds=json_dict["window_size_sec"]),
+        )
+
+
+class Ratio(Metric):
+    """
+    A Metric that shows the proportion of the number features that meets filter_expr
+    to the number of all features.
+    """
+
+    def __init__(
+        self,
+        filter_expr: str,
+        window_size: timedelta = timedelta(seconds=0),
+    ):
+        """
+        :param filter_expr: A partial FeatHub expression which evaluates to a boolean
+                            value. The partial Feathub expression should be a binary
+                            operator whose left child is absent and would be filled in
+                            with the host feature name. For example, "IS NULL" will be
+                            enriched into "{feature_name} IS NULL". Only features that
+                            evaluate this expression into True will be considered when
+                            computing the metric.
+        :param window_size: The time range to compute the metric. It should be zero or
+                            a positive time span. If it is zero, the metric will be
+                            computed from all features that have been processed since
+                            the Feathub job is created.
+        """
+        super().__init__("ratio", window_size)
+        self.filter_expr = filter_expr
+
+    def get_tags(self) -> OrderedDict:
+        tags = super(Ratio, self).get_tags()
+        tags["filter_expr"] = self.filter_expr
+        return tags
+
+    def get_transform_functions(self) -> Sequence[Callable[[str], Transformation]]:
+        return [
+            lambda feature_name: SlidingWindowTransform(
+                expr=f"CASE WHEN {feature_name} {self.filter_expr} "
+                f"THEN 1 ELSE 0 END",
+                agg_func="AVG",
+                window_size=self.window_size,
+                step_size=self.window_size,
+            ),
+            lambda feature_name: ExpressionTransform(
+                f"CASE WHEN {feature_name} IS NULL THEN 0 ELSE " f"{feature_name} END",
+            ),
+        ]
+
+    @append_metadata_to_json
+    def to_json(self) -> Dict:
+        return {
+            "window_size_sec": self.window_size / timedelta(seconds=1),
+            "filter_expr": self.filter_expr,
+        }
+
+    @classmethod
+    def from_json(cls, json_dict: Dict) -> "Ratio":
+        return Ratio(
+            filter_expr=json_dict["filter_expr"],
+            window_size=timedelta(seconds=json_dict["window_size_sec"]),
+        )
--- a/python/feathub/metric_stores/metric_store.py
+++ b/python/feathub/metric_stores/metric_store.py
@@ -1,0 +1,233 @@
+# Copyright 2022 The FeatHub Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import collections
+from abc import ABC, abstractmethod
+from datetime import timedelta
+from typing import Dict, List, Optional, OrderedDict, Tuple, Sequence
+
+from feathub.common.utils import generate_random_name
+from feathub.feature_tables.sinks.sink import Sink
+from feathub.feature_views.derived_feature_view import DerivedFeatureView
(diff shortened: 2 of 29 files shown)


yunfengzhou-hub:

python/feathub/metric_stores/metric_store.py
```diff
@@ -0,0 +1,86 @@
...
+            from feathub.metric_stores.prometheus_metric_store import (
+                PrometheusMetricStore,
+            )
+
+            return PrometheusMetricStore(props=props)
+
+        raise RuntimeError(f"Failed to instantiate metric store with props={props}.")
+
+    @property
+    def config(self) -> Dict:
```
lindong28, line 66: Would it be simpler to remove this since it is not used anywhere?

python/feathub/metric_stores/metric_store_config.py
```diff
@@ -0,0 +1,65 @@
...
+METRIC_STORE_PREFIX = "metric_store."
+
+METRIC_STORE_TYPE_CONFIG = METRIC_STORE_PREFIX + "type"
+METRIC_STORE_TYPE_DOC = "The type of the metric store to use."
+
+METRIC_STORE_REPORT_INTERVAL_CONFIG = METRIC_STORE_PREFIX + "report_interval"
+METRIC_STORE_REPORT_INTERVAL_DOC = "The interval to report metrics."
+
+METRIC_STORE_NAMESPACE_CONFIG = METRIC_STORE_PREFIX + "namespace"
+METRIC_STORE_NAMESPACE_DOC = "The namespace of the metric store."
```
lindong28, line 35: Should we document the impact of this config? For example, how does this config affects FeatHub output?

python/feathub/processors/flink/table_builder/prometheus_utils.py
```diff
@@ -0,0 +1,80 @@
...
+        if description:
+            builder._j_builder.withComment(description)
+
+    flink_schema = builder.build()
+
+    descriptor_builder = (
+        NativeFlinkTableDescriptor.for_connector("prometheus")
+        .schema(flink_schema)
+        .option("hostUrl", sink.host_url)
+        .option("jobName", generate_random_name("feathub_metrics"))
```
lindong28, line 62: How does jobName related to the namespace `metric_store.namespace`? It seems that they are conceptually the same thing? Can we keep only one of them?

python/feathub/processors/flink/table_builder/metric_store_utils.py
```diff
@@ -0,0 +1,151 @@
...
+from feathub.metric_stores.prometheus_metric_store import PrometheusMetricStore
+from feathub.processors.flink.table_builder.source_sink_utils_common import (
+    generate_random_name,
+)
+from feathub.processors.materialization_descriptor import MaterializationDescriptor
+from feathub.table.table_descriptor import TableDescriptor
+
+
+def _get_metric_name(metric: Metric, metric_store: MetricStore) -> str:
+    return f"feathub_{metric_store.namespace}_feature_{metric.metric_type}"
```
lindong28, line 39: Can you add doc explaining the expected output in Prometheus corresponding to arbitrary metric definition?

python/feathub/processors/flink/table_builder/metric_store_utils.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+def _get_metric_tags(metric: Metric, feature: Feature) -> typing.OrderedDict[str, str]:
+    tags = collections.OrderedDict(
+        [
+            ("feature_name", feature.name),
+            ("window_time", str(metric.window_size)),
+        ]
+    )
+    if isinstance(metric, Count) or isinstance(metric, Ratio):
```
lindong28, line 49: It seems more extensible to move the logic of determining the metric tags to the metric class rather than enumerating all metric subclasses in a util class here.

python/feathub/processors/flink/table_builder/metric_store_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    for feature in features_desc.get_output_features():
+        for metric in feature.metrics:
+            metric_tags = _get_metric_tags(metric, feature)
+            metric_tags["metric_name"] = _get_metric_name(metric, metric_store)
+            feature_name = generate_random_name("feathub_metric")
+            if isinstance(metric, Count):
+                metric_features.append(
+                    Feature(
+                        name=feature_name,
+                        transform=SlidingWindowTransform(
```
lindong28, line 76: Would it be more extensible to move the logic of converting metric definition into feature definition into the corresponding metric subclasses?

lindong28 (commented):

yunfengzhou-hub:
Hi @lindong28, thanks for the comments. I have updated the PR according to the comments.

python/feathub/metric_stores/metric.py
```diff
@@ -0,0 +1,210 @@
...
+        }
+
+    def get_agg_transform(
+        self, feature_name: str, step_size: timedelta
+    ) -> Union[str, Transformation]:
+        return SlidingWindowTransform(
+            expr="1",
+            agg_func="COUNT",
+            window_size=self.window_size,
+            step_size=step_size,
```
lindong28, line 122: Suppose window_size = 5 minutes and report_interval = 1 sec. This approach is going to use a sliding window with total of `5 min / 1 sec` = 300 steps. And the sliding window will need to keep track of up to 300 values internally.

Previously we discussed another approach where we can use one sliding window with window_size = step_size = 5 minute. And we can attach an extra sliding window with window_size = step_size = 1 sec AND agg_func = last. Then we will only need to maintain 2 values internally for this metric.

Also note that in general we don't expect report_interval to affect how the metric value is calculated. This current implementation breaks this expectation.

python/feathub/metric_stores/prometheus_metric_store.py
```diff
@@ -0,0 +1,164 @@
...
+    def _get_metrics_view(
+        self, features_desc: TableDescriptor, data_sink: Sink
+    ) -> TableDescriptor:
+        metric_features = []
+        post_process_metric_features = []
+        for feature in features_desc.get_output_features():
+            for metric in feature.metrics:
+                metric_tags = self._get_metric_tags(metric, feature, data_sink)
+                metric_tags["metric_name"] = self._get_metric_name(metric)
+                metric_feature_name = generate_random_name("feathub_metric")
```
lindong28, line 128: Would it be simpler to initialize the metric_feature_name according to the metric name pattern specified in `PrometheusMetricStore`'s python doc?

python/feathub/metric_stores/metric.py
```diff
@@ -0,0 +1,210 @@
...
+    def get_tags(self) -> Dict:
+        """
+        Gets the tags of this metric for tag-based metric reporters.
+        """
+        return {
+            "window_time": str(self.window_size),
+        }
+
+    @abstractmethod
+    def get_agg_transform(
```
lindong28, line 52: Would it be simpler not to have `agg` in the method name?

python/feathub/metric_stores/metric.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        Gets the transformation to aggregate and compute the metric value.
+
+        :param feature_name: The name of the host feature.
+        :param step_size: The time interval to get the aggregation result.
+        """
+        pass
+
+    @abstractmethod
+    def get_post_agg_transform(
```
lindong28, line 64: Would it be simpler and more extensible to have one method that returns a list of transformations instead of having two methods that return one transformation each?

python/feathub/metric_stores/metric.py
```diff
@@ -0,0 +1,210 @@
...
+        Gets the tags of this metric for tag-based metric reporters.
+        """
+        return {
+            "window_time": str(self.window_size),
+        }
+
+    @abstractmethod
+    def get_agg_transform(
+        self, feature_name: str, step_size: timedelta
+    ) -> Union[str, Transformation]:
```
lindong28, line 54: Would it be more readable to just have this method return `Transformation`?

Note that we use `str` for end-user to specify ExpressionTransform in a simpler way. But it might not be necessary to support the same shortcut for codes that are only used by FeatHub developers.

python/feathub/metric_stores/prometheus_metric_store.py
```diff
@@ -0,0 +1,164 @@
...
+        self, metric: Metric, feature: Feature, data_sink: Sink
+    ) -> Dict[str, str]:
+        return {
+            **metric.get_tags(),
+            "feature_name": feature.name,
+            "table_name": data_sink.name,
+        }
+
+    # TODO: support treating zero window as infinite window in SlidingWindowTransform
+    def _get_metrics_view(
```
lindong28, line 119: This method does not seem to depend on any thing specific to Prometheus. Would it be better to move this either to a shared util function or to the parent class `MetricStore`?

Same for other methods that are not specific to Prometheus.

python/feathub/processors/flink/flink_processor.py
```diff
@@ -233,6 +240,16 @@ def materialize_features(
                 # a local registry.
                 join_tables.update(self._get_join_tables(resolved_feature_descriptor))
 
+            metric_descriptor = None
+            if self.metric_store is not None:
+                metric_descriptor = self.metric_store.materialize_metrics_if_any(
```
lindong28, line 245: Most existing methods named `*materialize*` actually starts a job to ETL metrics. It is inconsistent and confusing to have a method named `materialize_metrics_if_any` that only creates the `MaterializationDescriptor`.

Would it be simpler to rename it as `create_metric_materialization_descriptor`?

python/feathub/metric_stores/metric_store.py
```diff
@@ -0,0 +1,92 @@
...
+        return timedelta(
+            seconds=self._metric_store_config.get(METRIC_STORE_REPORT_INTERVAL_CONFIG)
+        )
+
+    @abstractmethod
+    def materialize_metrics_if_any(
+        self,
+        feature_descriptor: TableDescriptor,
+        data_sink: Sink,
+    ) -> Optional[MaterializationDescriptor]:
```
lindong28, line 82: Is it sufficient to have this method return up to one `MaterializationDescriptor`?

What if user defined metrics with different window size for a given feature? 

We also need test to verify that users can define metrics with different window_size for the same feature. Feel free to add a TODO if it is not easy to do in this PR.

lindong28 (commented):

python/feathub/metric_stores/prometheus_metric_store.py
yunfengzhou-hub: The behavior of adding metrics tags as descriptions to features might be exclusive to prometheus store, or at least not suitable for metric stores without support for tags. Thus I would prefer to move common logic to `MetricStore` when we add local or other metric stores.

python/feathub/metric_stores/prometheus_metric_store.py
yunfengzhou-hub: Given that each metric need to generate a sequence of transformations and we need columns to store the intermediate result, and that metrics might have identical metric names, the practice of generating random metric name is still kept in the latest implementation.

python/feathub/metric_stores/metric.py
yunfengzhou-hub: I have implemented the behavior of window size = step size = 5 min.

Further appending a next sliding window to support report interval caused the following exception in tests: 
```
java.lang.RuntimeException: Record has Long.MIN_VALUE timestamp (= no timestamp marker). Is the time characteristic set to 'ProcessingTime', or did you forget to call 'DataStream.assignTimestampsAndWatermarks(...)'?
```
It seems that the upstream sliding window has lost timestamp attribute somewhere, but I haven't found the bug in code. I have added a TODO for supporting report interval.

yunfengzhou-hub:

python/feathub/metric_stores/prometheus_metric_store.py
```diff
@@ -0,0 +1,220 @@
...
+                        metric_name
+                    ).description,
+                )
+                for metric_name in metric_name_count_dict.keys()
+            ],
+            keep_source_fields=False,
+        )
+
+    def _escape_tag_value_str(self, tag_value: str) -> str:
+        characters_to_escape = [",", ";", "="]
```
lindong28, line 209: Given that there is no known use-case for using these characters in the metric label, would it be simpler to just throw exception the `tag_value` contains these characters?

python/feathub/metric_stores/prometheus_metric_store.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+# TODO: Allow users to configure feature name pattern.
+class PrometheusMetricStore(MetricStore):
+    """
+    A metric store reporting metrics to Prometheus through push gateway. For
+    user-defined Metrics, they would be reported in the following format:
+
+    - metric name: "{namespace}_{feature_name}_{metric_type}".
+    - tags: the metric tags along with the following tags.
```
lindong28, line 74: Since Prometheus use `label` instead of `tag`, should we use label as appropriate for Prometheus-specific classes?

lindong28 (commented):

python/feathub/metric_stores/metric.py
lindong28: If it is not quick to fix, it might be simpler to add a UDF simply cache the last input value and periodically emits the value using processing time timer.

python/feathub/metric_stores/metric.py
lindong28: It seems that `get_transform_functions()` still uses `Union[str, Transformation]` in the method signature.

Would it be simpler for this method to return `Sequence[Callable[[str], Transformation]]`?

lindong28: It seems that `get_transform_functions()` still uses `Union[str, Transformation]` in the method signature.

Would it be simpler for this method to return `Sequence[Callable[[str], Transformation]]`?

python/feathub/metric_stores/prometheus_metric_store.py
lindong28: All metric store to be supported by FeatHub should have the `tags` concepts. Otherwise, we would not have the API `Metric#get_tags`. It would be very surprising for a popular metric store not to support `tag` for metrics.

In addition, it seems that the most reasonable way to pass label to the underlying metric store is via the feature's description field, which is exactly what `_get_metrics_view()` does. So it seems the logic in this method is reusable by other metric store.

yunfengzhou-hub: I double-checked Flink's metric reporters and found that they all support tags, so I generally agree with this idea. Some details that differs prometheus/local metric store are as follows.

- Prometheus sink supports using `extraLabels` to configure common labels for all metrics in a feature view. Thus while the default `_get_metric_tags` in `MetricStore` need to specify tags like `table_name` in `_get_metric_tags`, PrometheusMetricStore need not.
- Prometheus uses the label `job` to stands for the `namespace` configuration, while others would still use the label name `namespace` by default.
- LocalMestricStore might use print sink or filesystem sink, neither of which supports parsing Table column comments as labels. Thus this metric store would need to add an extra DerivedFeatureView to merge the comments/descriptions back to metric value before sinking.

I have updated metric_store.py and prometheus_metric_store.py with the above details.

python/feathub/metric_stores/prometheus_metric_store.py
yunfengzhou-hub: For now I agree with it that there is no known use-case for these characters, so I'll remove this method in this PR. In the long run there still might be such use cases. Suppose a user wants to compute the ratio of incorrect data in the features, and the criteria is the hash of the feature's value. In this case the label string passed to prometheus sink might look like follows.
```
...feature_name=xxx,filter_expr=digest=SHA2(feature_a, 256),window_size=3600...
```
Here the `,` and `=` might cause exceptions when the sink is parsing this string.

python/feathub/metric_stores/metric.py
yunfengzhou-hub: `PythonUDFTransform` only supports simple mapping functions without states or timer service, so it might not be simple to achieve the UDF described above with existing infrastructure.

Sxnan (commented):
Thanks for the PR! The PR LGTM overall. Just left one comment below.

python/feathub/metric_stores/prometheus_metric_store.py
Sxnan: I think there are use-case that use these characters. 
```python
Count(
    filter_expr="= 'a,b'",
    window_size=timedelta(days=1),
) 
```I think there are use-case that use these characters. 
```python
Count(
    filter_expr="= 'a,b'",
    window_size=timedelta(days=1),
) 
```

yunfengzhou-hub: Yes you are correct. There are known use cases for now. I'll keep this method.

Sxnan:
@yunfengzhou-hub Thanks for the update. LGTM! @lindong28 Can you have another look at this PR?

python/feathub/metric_stores/metric.py
lindong28: Is it feasible to use Java UDF?

python/feathub/metric_stores/prometheus_metric_store.py
lindong28: Sounds good.

python/feathub/metric_stores/prometheus_metric_store.py
lindong28: Sounds good.

lindong28:
Thanks for the update. LGTM.

PR #171, opened 2023-07-27
Add document and support report interval in metric store

yunfengzhou-hub:

This pull request adds document and support for report interval in metric store.

## Brief change log

- Add document for metric store and metrics.
- Add JavaUdfTransform
- Support report interval in metric store through JavaUdfTransform.

  - The public API: no

## Documentation

### DIFF
--- a/docs/content/metric-stores/overview.md
+++ b/docs/content/metric-stores/overview.md
@@ -1,0 +1,113 @@
+# Overview
+
+A MetricStore provides properties to set the metrics of a Feathub job into an
+external metric service. A metric refers to a statistic of a characteristic of a
+feature.
+
+## Metric store configurations
+
+Below are common configurations shared by different metric store
+implementations. The document of each metric store contains its specific
+configurations.
+
+| Key                 | Required | Default   | Type    | Description                                                  |
+| ------------------- | -------- | --------- | ------- | ------------------------------------------------------------ |
+| type                | Required | -         | String  | The type of the metric store to use.                         |
+| report_interval_sec | Optional | 10        | Float   | The interval in seconds to report metrics.                   |
+| namespace           | Optional | "default" | String  | The namespace to report metrics to the metric store. Metrics within different namespace will not overwrite each other. |
+
+## Defining metrics
+
+Feathub supports defining metrics at feature's granularity. Below is an example
+of defining a metric for a feature.
+
+```python
+f_total_cost = Feature(
+    name="total_cost",
+    transform=SlidingWindowTransform(
+        expr="cost",
+        agg_func="SUM",
+        group_by_keys=["name"],
+        window_size=timedelta(days=2),
+        step_size=timedelta(days=1),
+    ),
+    metrics=[
+    	Count(
+          filter_expr="> 100",
+          window_size=timedelta(days=1),
+      ),
+    	Ratio(
+          filter_expr="IS NULL",
+          window_size=timedelta(hours=1),
+      ),
+    ],
+)
+```
+
+## Metric reporting format
+
+Metrics reported by metric stores will have the following format by default.
+Some metric stores might override the default format, and please check the
+document of each metric store for details.
+
+- Metric name: `"{namespace}_{feature_name}_{metric_type}"`
+  - namespace: The namespace of the metric store.
+  - feature_name: The name of the host feature.
+  - metric_type: The type of the metric.
+- Metric tags:
+  - namespace: The namespace of the metric store.
+  - table_name: The name of the sink where the host features would be written
+    to.
+  - feature_name: The name of the host feature.
+  - other metric-specific tags.
+
+## Built-in metrics
+
+Below are Feathub's built-in metrics's metric types, their parameters and their
+exposed tags.
+
+### Count
+
+Count is a metric that shows the number of features. It has the following
+parameters:
+
+- filter_expr: Optional with None as the default value. If it is not None, it
+  represents a partial FeatHub expression which evaluates to a boolean value.
+  The partial Feathub expression should be a binary operator whose left child is
+  absent and would be filled in with the host feature name. For example, "IS
+  NULL" will be enriched into "{feature_name} IS NULL". Only features that
+  evaluate this expression into True will be considered when computing the
+  metric.
+- window_size: Optional with 0 as the default value. The time range to compute
+  the metric. It should be zero or a positive time span. If it is zero, the
+  metric will be computed from all feature values that have been processed since
+  the Feathub job is created.
+
+It exposes the following metric-specific tags:
+
+- metric_type: "count"
+- filter_expr: The value of the filter_expr parameter.
+- window_size_sec: The value of the window_size parameter in seconds.
+
+### Ratio
+
+Ratio is a metric that shows the proportion of the number features that meets
+filter_expr to the number of all features. It has the following parameters:
+
+- filter_expr: A partial FeatHub expression which evaluates to a boolean value.
+  The partial Feathub expression should be a binary operator whose left child is
+  absent and would be filled in with the host feature name. For example, "IS
+  NULL" will be enriched into "{feature_name} IS NULL". Only features that
+  evaluate this expression into True will be considered when computing the
+  metric.
+- window_size: Optional with 0 as the default value. The time range to compute
+  the metric. It should be zero or a positive time span. If it is zero, the
+  metric will be computed from all feature values that have been processed since
+  the Feathub job is created.
+
+It exposes the following metric-specific tags:
+
+- metric_type: "ratio"
+- filter_expr: The value of the filter_expr parameter.
+- window_size_sec: The value of the window_size parameter in seconds.
+
--- a/docs/content/metric-stores/prometheus.md
+++ b/docs/content/metric-stores/prometheus.md
@@ -1,0 +1,62 @@
+# Prometheus
+
+The PrometheusMetricStore supports reporting metrics to a Prometheus service
+through [Prometheus Pushgateway](https://github.com/prometheus/pushgateway). 
+
+## Metric reporting format
+
+The metrics reported by PrometheusMetricStore generally match the default format
+with the following modifications.
+
+- It uses the label name "job" instead of "namespace" to hold the namespace
+  value of the metric store.
+- table_name is not only reported as a label of the metric, but also the
+  grouping key of the job in Prometheus PushGateway.
+
+
+## Configurations
+
+The following describes the configurations required to set up a
+PrometheusMetricStore instance.
+
+| Key                | Required | Default | Type    | Description                                                  |
+| ------------------ | -------- | ------- | ------- | ------------------------------------------------------------ |
+| server_url         | Required | -       | String  | The PushGateway server URL including scheme, host name, and port. |
+| delete_on_shutdown | Optional | True    | Boolean | Whether to delete metrics from Prometheus when the job finishes. When set to true, Feathub will try its best to delete the metrics but this is not guaranteed. |
+
+## Examples
+
+Here is an example that creates a FeathubClient that reports metrics to
+Prometheus.
+
+```python
+client = FeathubClient(
+    {
+        "processor": {
+            "type": "local",
+        },
+        "online_store": {
+            "types": ["memory"],
+            "memory": {},
+        },
+        "registry": {
+            "type": "local",
+            "local": {
+                "namespace": "default",
+            },
+        },
+        "feature_service": {
+            "type": "local",
+            "local": {},
+        },
+      	"metric_store": {
+            "type": "prometheus",
+            "report_interval_sec": 5,
+            "prometheus": {
+                "server_url": "localhost:8080",
+                "delete_on_shutdown": False,
+            },
+        }
+    }
+)
+```
--- a/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
+++ b/java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2022 The FeatHub Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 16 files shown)


yunfengzhou-hub:
Hi @Sxnan could you please take a look at this PR?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PeriodicEmitLastValueOperator.java
```diff
@@ -0,0 +1,81 @@
...
+    public void initializeState(StateInitializationContext context) throws Exception {
+        super.initializeState(context);
+        state =
+                context.getOperatorStateStore()
+                        .getListState(new ListStateDescriptor<>("lastRowState", Row.class));
+    }
+
+    @Override
+    public void processElement(StreamRecord<Row> element) throws Exception {
+        if (!state.get().iterator().hasNext()) {
```
Sxnan, line 54: This introduces one state access for each processElement call. Can we avoid the state access by introducing a flag in memory to keep track of whether the state is empty? 

And we can initialize the flag during state initialization.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PeriodicEmitLastValueOperator.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    private class EmitLastValueCallback implements ProcessingTimeService.ProcessingTimeCallback {
+        @Override
+        public void onProcessingTime(long l) throws Exception {
+            output.collect(new StreamRecord<>(state.get().iterator().next()));
+            getProcessingTimeService()
+                    .registerTimer(
+                            getProcessingTimeService().getCurrentProcessingTime() + emitIntervalMs,
```
Sxnan, line 77: I think we can use is passed as parameter `l` to register the next processing time timer.

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -39,8 +40,8 @@ class DerivedFeatureView(FeatureView):
     """
     Derives features by applying the given transformations on an existing table.
 
-    Supports per-row transformation, over window aggregation transformation and table
-    join.
+    Supports per-row transformation, over window aggregation transformation, table
+    join and Java UDF transformation.
```
Sxnan, line 44: I think we need more discussion before exposing the Java UDF transform. Should we remove the docs in this PR? And we can support Java UDF transformation properly in a future PR.

python/feathub/metric_stores/metric_store.py
```diff
@@ -222,6 +222,27 @@ def _get_metrics_view(
...
 
+        schema_builder = Schema.new_builder()
+        for feature in derived_feature_view.get_output_features():
+            schema_builder.column(feature.name, feature.dtype)
+
+        return DerivedFeatureView(
+            name=f"{features_desc.name}_periodic_reported_metrics_{window_size_sec}",
+            source=derived_feature_view,
+            features=[
+                Feature(
```
Sxnan, line 233: However, In the long-term, I think it makes more sense to put the JavaUdfTransform, which is a row-level transformation, in the FeatureView, similar to the filter expr in FeatureView.

Sxnan (commented):
Thanks for the PR. I left some comments below.

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
```diff
@@ -0,0 +1,57 @@
...
+     */
+    @SuppressWarnings({"unchecked"})
+    public static DataStream<Row> evalJavaUdf(
+            DataStream<Row> stream, String className, List<?> parameters) throws Exception {
+        // TODO: Support configuring the type information of the output stream.
+        return stream.transform(
+                        "JavaUDF",
+                        stream.getType(),
+                        (OneInputStreamOperator<Row, Row>) instantiate(className, parameters))
+                .setParallelism(1);
```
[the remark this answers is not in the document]
yunfengzhou-hub: According to offline discussion, this PR would focus on supporting the report interval function of metric store. Generalizing this function into a public API requires more workload, which is beyond the scope of this PR. Same for other comments.

docs/content/metric-stores/prometheus.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Configurations
+
+The following describes the configurations required to set up a
+PrometheusMetricStore instance.
+
+| Key                | Required | Default | Type    | Description                                                  |
+| ------------------ | -------- | ------- | ------- | ------------------------------------------------------------ |
+| server_url         | Required | -       | String  | The PushGateway server URL including scheme, host name, and port. |
+| delete_on_shutdown | Required | True    | Boolean | Whether to delete metrics from Prometheus when the job finishes. When set to true, Feathub will try its best to delete the metrics but this is not guaranteed. |
```
lindong28, line 25: Required -> Optional

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    @SuppressWarnings({"unchecked"})
+    public static DataStream<Row> evalJavaUdf(
+            DataStream<Row> stream, String className, List<?> parameters) throws Exception {
+        // TODO: Support cases in which the input and output type differs.
+        return stream.transform(
+                        "JavaUDF",
+                        stream.getType(),
+                        (OneInputStreamOperator<Row, Row>) instantiate(className, parameters))
+                // TODO: Generify the configuration of parallelism.
```
lindong28, line 45: Rate and Count metrics are both computed using `SlidingWindowTransform` with `group_by_keys = ()`, which the corresponding DataStream should already have parallelism = 1.

Instead of giving developers the idea that we are changing the parallelism to 1 here, it might be more readable to verify that the input `stream` already has parallelism=1.

In the future, when we have metrics whose DataStream can have parallelism > 1, we might find it reasonable to set the parallelism to be same as `stream` here.

lindong28 (commented):

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,106 @@
...
+  with the host feature name. For example, "IS NULL" will be enriched into
+  "{feature_name} IS NULL". Only features that evaluate this expression into
+  True will be considered when computing the metric.
+- window_size: The time range to compute the metric. It should be zero or a
+  positive time span. If it is zero, the metric will be computed from all
+  features that have been processed since the Feathub job is created.
+
+It exposes the following tags:
+
+- filter_expr: The value of the filter_expr parameter.
```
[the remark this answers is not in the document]
yunfengzhou-hub: I will a note in `Metric reporting format` section instead of in the section for each metric.

Besides, PrometheusSink has been able to identify the escaped characters and de-escape them. Do you think other metric stores will also have this ability, and finally the escaping behavior would only exist in internal implementation and not exposed to users?

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
yunfengzhou-hub: When SlidingWindowTransform has group_by_keys = (), the current implementation would still generate a result datastream whose parallelism is larger than 1. This is also reasonable given that records with the same key can be pre-aggregated on different subtasks at first for better performance, according to the JavaDoc of `org.apache.flink.api.common.functions.AggregateFunction`.

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,106 @@
...
+## Metric store configurations
+
+Below are common configurations shared by different metric store
+implementations. The document of each metric store contains its specific
+configurations.
+
+| Key                 | Required | Default   | Type    | Description                                                  |
+| ------------------- | -------- | --------- | ------- | ------------------------------------------------------------ |
+| type                | Required | -         | String  | The type of the metric store to use.                         |
+| report_interval_sec | Required | -         | Integer | The interval in seconds to report metrics.                   |
```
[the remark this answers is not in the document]
yunfengzhou-hub: I have changed the type of this configuration to `Number` here. 

This configurations definition and the configuration validation process is also updated to support `Union[float, int]`. Please refer to the latest code for details.

yunfengzhou-hub:

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/JavaUdfUtils.java
lindong28: Would it be more readable to verify that the input stream already has parallelism=1?

docs/content/metric-stores/overview.md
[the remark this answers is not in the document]
lindong28: Given that Prometheus will show the label value as if it has not been escaped, I agree it is reasonable not to document how we escape the label value.

docs/content/metric-stores/overview.md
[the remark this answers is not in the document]
lindong28: `Float` is simpler to understand than `Number`. For example, `jobmanager.memory.jvm-overhead.fraction` has type `Float` in Flink doc [1].

Can you explain why we should use `Union[float, int]` instead of simply `float`?

[1] https://nightlies.apache.org/flink/flink-docs-master/docs/deployment/config/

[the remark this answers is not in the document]
yunfengzhou-hub: If the type of this configuration is `float` and users configure`report_interval_sec: 5`, an exception would be raised here.
https://github.com/alibaba/feathub/blob/master/python/feathub/common/config.py#L178

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,110 @@
...
+## Metric store configurations
+
+Below are common configurations shared by different metric store
+implementations. The document of each metric store contains its specific
+configurations.
+
+| Key                 | Required | Default   | Type    | Description                                                  |
+| ------------------- | -------- | --------- | ------- | ------------------------------------------------------------ |
+| type                | Required | -         | String  | The type of the metric store to use.                         |
+| report_interval_sec | Required | -         | Float   | The interval in seconds to report metrics.                   |
```
lindong28, line 16: How about setting the default value to 10 sec by default?

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,110 @@
...
+
+Below are Feathub's built-in metrics's metric types, their parameters and their
+exposed tags.
+
+### Count
+
+Count is a metric that shows the number of features. It has the following
+parameters:
+
+- filter_expr: Optional. If it is not None, it represents a partial FeatHub
```
lindong28, line 74: Since this is optional, can you also specify the default value?

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,110 @@
...
+  expression which evaluates to a boolean value. The partial Feathub expression
+  should be a binary operator whose left child is absent and would be filled in
+  with the host feature name. For example, "IS NULL" will be enriched into
+  "{feature_name} IS NULL". Only features that evaluate this expression into
+  True will be considered when computing the metric.
+- window_size: The time range to compute the metric. It should be zero or a
```
lindong28, line 80: Would it be useful to make it optional with 0 as the default value?

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,110 @@
...
+  positive time span. If it is zero, the metric will be computed from all
```
lindong28, line 81: "all features" -> "all feature values"

java/feathub-udf/flink-udf/src/main/java/com/alibaba/feathub/flink/udf/PeriodicEmitLastValueOperator.java
```diff
@@ -0,0 +1,90 @@
...
+
+import java.io.Serializable;
+import java.util.Collections;
+import java.util.Iterator;
+
+/** An operator that periodically emits the last value it has received from upstream operator. */
+public class PeriodicEmitLastValueOperator extends AbstractStreamOperator<Row>
+        implements OneInputStreamOperator<Row, Row>, Serializable, BoundedOneInput {
+    private final long emitIntervalMs;
+    private Row lastValue;
```
lindong28, line 38: Can we mark this variable nullable?

python/feathub/metric_stores/metric_store.py
```diff
@@ -47,6 +51,14 @@ def __init__(self, config: Dict) -> None:
         """
         self._metric_store_config = MetricStoreConfig(config)
 
+        # TODO: Add validators to verify the precision of configurations
+        #  and validators to express and/or logic.
+        interval = self._metric_store_config.get(
+            METRIC_STORE_REPORT_INTERVAL_SEC_CONFIG
+        )
+        if not interval.is_integer():
```
lindong28, line 59: Do we still need this check?

lindong28 (commented):

python/feathub/metric_stores/metric_store.py
yunfengzhou-hub: Given that Feathub is supposed to support intervals whose precision is at most in seconds, this code helps to identify invalid configurations like `report_interval_sec = 5.5`.

yunfengzhou-hub:

python/feathub/metric_stores/metric_store.py
lindong28: Based on the offline discussion, we have agreed to remove this check.

docs/content/metric-stores/overview.md
```diff
@@ -0,0 +1,113 @@
...
+Ratio is a metric that shows the proportion of the number features that meets
+filter_expr to the number of all features. It has the following parameters:
+
+- filter_expr: A partial FeatHub expression which evaluates to a boolean value.
+  The partial Feathub expression should be a binary operator whose left child is
+  absent and would be filled in with the host feature name. For example, "IS
+  NULL" will be enriched into "{feature_name} IS NULL". Only features that
+  evaluate this expression into True will be considered when computing the
+  metric.
+- window_size: The time range to compute the metric. Optional with 0 as the
```
lindong28, line 103: Should we put `Optional with 0 as the default value` right after  `window_size` for consistency with `filter_expr`'s doc?

lindong28:
Thanks for the PR. LGTM.

We can merge the PR after CI passes.

Issue #170, opened 2023-07-26
Avoid join operation for SlidingWindowTransforms with different filters


yunfengzhou-hub:
In current FlinkProcessor's implementation, If a SlidingFeatureView contains SlidingWindowTransforms with different filter expressions, multiple Flink windowing operation would be applied and the aggregation results would be joined to make the final table. This could be optimized so that for each SlidingFeatureView, there is only 1 windowing operation and 0 join operation.

PR #173, opened 2023-07-31
Support computing sliding window features with different filter expr and limits in one operator

### LINKED ISSUE #97: Remove join operation between window transforms with different limit parameter
Currently in FlinkProcessor, SlidingWindowTransforms and OverWindowTransforms whose `limit` parameters are different would be evaluated separately and joined together afterward, which has added to the performance overhead in some feature-generation cases. It should be optimized to evaluate transformations matching this character with the same process function, and remove the unnecessary joins.

Sxnan:

Support computing sliding window features with different filter expr and limits in one operator to avoid full outer join after sliding window.

Fix #97 #170 

## Brief change log
- Support computing sliding window features with different filter expr and limits in one operator

  - The public API: no

## Documentation

### DIFF
--- a/python/feathub/feature_views/sliding_feature_view.py
+++ b/python/feathub/feature_views/sliding_feature_view.py
@@ -11,6 +11,7 @@
 #  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
+from datetime import timedelta
 from typing import Dict, Union, Sequence, Set, Any, Optional, List, OrderedDict, cast
 
 from feathub.common import types
@@ -186,6 +187,9 @@ class SlidingFeatureView(FeatureView):
                 "is forbidden."
             )
 
+        self.step_size = self._get_step_size()
+        self.group_by_keys = self._get_group_by_keys()
+
     @staticmethod
     def _get_window_time_feature(
         timestamp_field: str, timestamp_format: str
@@ -269,6 +273,59 @@ class SlidingFeatureView(FeatureView):
             extra_props={**props, **self.config.original_props},
         )
         return feature_view
+
+    def _get_step_size(self) -> timedelta:
+        """
+        Returns the step size of the sliding window features in the feature view.
+        """
+        step_size_set = set(
+            [
+                feature.transform.step_size
+                for feature in self.features
+                if isinstance(feature, Feature)
+                and isinstance(feature.transform, SlidingWindowTransform)
+            ]
+        )
+
+        if len(step_size_set) < 1:
+            raise FeathubException(
+                "SlidingWindowFeatureView must have at least one feature with "
+                "SlidingWindowTransform."
+            )
+
+        if len(step_size_set) > 1:
+            raise FeathubException(
+                f"SlidingWindowTransforms have different step size " f"{step_size_set}."
+            )
+
+        return step_size_set.pop()
+
+    def _get_group_by_keys(self) -> Sequence[str]:
+        """
+        Returns the group-by keys of the sliding window features in the feature view.
+        """
+        group_by_keys_set = set(
+            [
+                tuple(feature.transform.group_by_keys)
+                for feature in self.features
+                if isinstance(feature, Feature)
+                and isinstance(feature.transform, SlidingWindowTransform)
+            ]
+        )
+
+        if len(group_by_keys_set) < 1:
+            raise FeathubException(
+                "SlidingWindowFeatureView must have at least one feature with "
+                "SlidingWindowTransform."
+            )
+
+        if len(group_by_keys_set) > 1:
+            raise FeathubException(
+                f"SlidingWindowTransforms have different group-by keys "
+                f"{group_by_keys_set}."
+            )
+
+        return group_by_keys_set.pop()
 
     @append_metadata_to_json
     def to_json(self) -> Dict:
@@ -371,11 +428,6 @@ class SlidingFeatureView(FeatureView):
         pre_sliding_features: Set[Feature],
         sliding_features: Set[Feature],
     ) -> None:
-        if len(sliding_features) < 1:
-            raise FeathubException(
-                "SlidingWindowFeatureView must have at least one feature with "
-                "SlidingWindowTransform."
-            )
 
         valid_variables = {
             *[f.name for f in source.get_output_features()],
@@ -411,17 +463,6 @@ class SlidingFeatureView(FeatureView):
         group_by_keys_set = set(
             [tuple(transform.group_by_keys) for transform in sliding_window_transforms]
         )
-        if len(group_by_keys_set) > 1:
-            raise FeathubException(
-                f"SlidingWindowTransforms have different group-by keys "
-                f"{group_by_keys_set}."
-            )
-
-        steps = set([transform.step_size for transform in sliding_window_transforms])
-        if len(steps) > 1:
-            raise FeathubException(
-                f"SlidingWindowTransforms have different step size " f"{steps}."
-            )
 
         # Validate that the per-row transform features are used group-by-keys of
         # SlidingWindowTransform.
--- a/python/feathub/processors/flink/table_builder/aggregation_utils.py
+++ b/python/feathub/processors/flink/table_builder/aggregation_utils.py
@@ -12,7 +12,7 @@
 #  See the License for the specific language governing permissions and
 #  limitations under the License.
 from datetime import timedelta
-from typing import Any, Tuple
+from typing import Any, Tuple, Optional
 
 from pyflink.table.types import DataType, DataTypes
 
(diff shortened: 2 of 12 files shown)


python/feathub/processors/flink/table_builder/aggregation_utils.py
```diff
@@ -41,12 +41,16 @@ def __init__(
         expr: str,
         agg_func: AggFunc,
         window_size: timedelta,
+        limit: int,
+        filter_expr: str,
```
yunfengzhou-hub, line 45: Given that this file is used in both sliding windows and over windows, it might be better to add some description or check that over window is not using these fields.

python/feathub/processors/flink/table_builder/sliding_window_utils.py
```diff
@@ -49,29 +46,17 @@ class SlidingWindowDescriptor:
...
-        self.limit = limit
         self.group_by_keys = group_by_keys
-        self.filter_expr = filter_expr
 
     @staticmethod
     def from_sliding_window_transform(
         sliding_window_agg: SlidingWindowTransform,
     ) -> "SlidingWindowDescriptor":
-        filter_expr = (
-            to_flink_sql_expr(sliding_window_agg.filter_expr)
```
yunfengzhou-hub, line 66: It seems that the feathub filter_expr instead of flink filter_expr is passed to java in this PR. Let's fix this bug and add a test for cases where these two filter_expr differs.

python/feathub/processors/flink/table_builder/flink_table_builder.py
```diff
@@ -536,46 +535,40 @@ def _get_table_from_sliding_feature_view(
...
+                    sliding_window_descriptor = (
+                        SlidingWindowDescriptor.from_sliding_window_transform(transform)
+                    )
+                elif (
+                    sliding_window_descriptor
+                    != SlidingWindowDescriptor.from_sliding_window_transform(transform)
+                ):
+                    raise FeathubException(
+                        "There can be only one sliding window descriptor for one "
+                        "SlidingFeatureView."
```
yunfengzhou-hub, line 548: Let's improve this exception message for Feathub users. For example, "a SlidingFeatureView cannot have different keys or step size".

Given that this requirement has existed before this PR, it should have been covered by existing exceptions. Could you please verify this? Perhaps the condition
```python
elif (
    sliding_window_descriptor
    != SlidingWindowDescriptor.from_sliding_window_transform(transform)
)
```
Would never happen.

yunfengzhou-hub (commented):
Thanks for the PR @Sxnan. Left some comments as below.

python/feathub/processors/flink/table_builder/flink_table_builder.py
Sxnan: It should never reach here. I updated SlidingFeatureView to let it report the group-by keys and step size, which is more intuitive IMO.

python/feathub/processors/flink/table_builder/aggregation_utils.py
Sxnan: I added a TODO to optimize the over window to support different window sizes, limit, and filter expr.

Sxnan:
@yunfengzhou-hub Thanks for your review! I updated the PR. Could you take another look?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,6 +271,50 @@ def build(
         )
         return feature_view
 
+    def get_step_size(self) -> timedelta:
```
yunfengzhou-hub, line 274: It might be better to declare this method as `@property`. Same for `get_group_by_keys`.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,6 +271,50 @@ def build(
...
+        """
+        Returns the step size of the sliding window features in the feature view. All
+        the sliding window features must have the same step size.
```
yunfengzhou-hub, line 277: It has been recorded in SlidingFeatureView's python doc that "SlidingWindowTransform used in this feature view must have the same step size and group-by keys", so it might be unnecessary to mention this requirement again here.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,6 +271,50 @@ def build(
...
+        """
+        step_size_set = set(
+            [
+                feature.transform.step_size
+                for feature in self.get_resolved_features()
+                if isinstance(feature.transform, SlidingWindowTransform)
+            ]
+        )
+        if len(step_size_set) == 0:
+            raise FeathubException("There must be at least one sliding window feature.")
```
yunfengzhou-hub, line 287: It might be better to move the checks to the constructor of SlidingFeatureView, so users can get the exception immediately after the class is created.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,6 +271,50 @@ def build(
...
 
+    def get_step_size(self) -> timedelta:
+        """
+        Returns the step size of the sliding window features in the feature view. All
+        the sliding window features must have the same step size.
+        """
+        step_size_set = set(
+            [
+                feature.transform.step_size
+                for feature in self.get_resolved_features()
```
yunfengzhou-hub, line 282: It seems that getting the step size of a sliding feature view does not necessarily require that the feature view is resolved, as it does not rely on source table's properties, nor would Feathub convert any `str` feature to sliding window transform.

yunfengzhou-hub (commented):
Thanks for the update. LGTM overall. Left some small comments as below.

@lindong28 could you please take a look at this PR?

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -270,6 +307,38 @@ def build(
...
+        step_size_set = set(
+            [
+                feature.transform.step_size
+                for feature in self.features
+                if isinstance(feature, Feature)
+                and isinstance(feature.transform, SlidingWindowTransform)
+            ]
+        )
+
+        return step_size_set.pop()
```
lindong28, line 324: The code is not very self-explanatory as it assumes that the set size == 1, and this assumption is verified in the constructor.

Can we either cache the step_size obtained in the constructor, or verify set size == 1 in this method?

Same for group_by_keys.

python/feathub/feature_views/sliding_feature_view.py
```diff
@@ -186,6 +187,42 @@ def __init__(
...
             )
 
+        sliding_features = [
+            feature
+            for feature in features
+            if isinstance(feature, Feature)
+            and isinstance(feature.transform, SlidingWindowTransform)
+        ]
+
+        if len(sliding_features) < 1:
```
lindong28, line 197: Would it be more readable to move these validation logic into the existing `_validate(...)` method?

python/feathub/processors/flink/table_builder/aggregation_utils.py
lindong28, line 45: Given that filter_expr can be none, would it be more readable to make it `Optional[str]`?

python/feathub/processors/flink/table_builder/aggregation_utils.py
```diff
@@ -41,12 +41,16 @@ def __init__(
         expr: str,
         agg_func: AggFunc,
         window_size: timedelta,
+        limit: int,
```
lindong28, line 44: Given that limit comes from transform.limit and its value can be none, should we make it `Optional[int]`?

python/feathub/processors/flink/table_builder/tests/test_aggregation_utils.py
```diff
@@ -0,0 +1,86 @@
...
+from feathub.feature_views.transforms.over_window_transform import OverWindowTransform
+from feathub.feature_views.transforms.sliding_window_transform import (
+    SlidingWindowTransform,
+)
+from feathub.processors.flink.table_builder.aggregation_utils import (
+    AggregationFieldDescriptor,
+)
+
+
+class AggregationUtilsTest(unittest.TestCase):
```
lindong28, line 31: Can you also add an integration test shows the benefit of this change for a Python program? For example, we should be able to show that when filter expression is used with OverWindow, the output column will no longer contain null value for rows that do not match the filter expression.

There is a request for this feature in the DingDing user group. I will copy/paste the message from user below:

```
1.譬如算 '用户最近1小时点击次数'， 
Feature(
        name="click_count",
        transform=OverWindowTransform(
            expr="bhv_type",
            agg_func="COUNT",
            window_size=timedelta(hours=1),
            group_by_keys=["user_id"],
            limit=None,
            filter_expr="bhv_type='click'"
        ),
    )
在 bhv_type不等于click时，赋null值，如果 该用户最近1小时有发生点击，则取null值不符合该特征的含义。
2.另外 A. group_by_keys 在窗口范围内没有事件会取null 值; B. filter_expr not match 取null值，这两种情况都会取 null值，导致没法区分

```

lindong28 (commented):

python/feathub/feature_views/sliding_feature_view.py
Sxnan: The `_validate` method should be invoked during build processing after all features are resolved. 

I updated the SlidingFeatureView so that the step_size and group_by_keys are computed and cached in the constructor. And the validation logic happens while computing the step_size and group_by_keys.

python/feathub/processors/flink/table_builder/tests/test_aggregation_utils.py
lindong28: Sounds good.

Sxnan:
@lindong28 Thanks for your review! I updated the PR. Could you take another look?

lindong28:
Thanks for the update. LGTM.

PR #176, opened 2023-08-03
Support reporting metrics in non-final feature view

yunfengzhou-hub:

This pull request adds support for reporting metrics defined in non-final feature view.

## Brief change log

  - Adds support for reporting metrics defined in non-final feature view.
  - Adds debug log when constructing metric reporting workflow during compilation.

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? docs & python docs

### DIFF
--- a/docs/content/metric-stores/overview.md
+++ b/docs/content/metric-stores/overview.md
@@ -44,6 +44,37 @@ f_total_cost = Feature(
 )
 ```
 
+Then when the FeatureView that hosts the above feature is being materialized,
+the related metrics will also be reported during the materialization process.
+
+By default, Feathub would only report the metrics directly defined in the
+FeatureView to be materialized. If the `keep_source_metrics` parameter is
+enabled when creating the feature view as follows, FeatHub will recursively
+enumerate the source feature view of this and every upstream feature view whose
+`keep_source_fields == true`, and report metrics defined in those feature views.
+
+```python
+feature_view_0 = ...
+
+feature_view_1 = SlidingFeatureView(
+    name="feature_view_1",
+    source=feature_view_0,
+    features=[f_total_cost], # A feature containg metrics.
+)
+
+feature_view_2 = DerivedFeatureView(
+    name="feature_view_2",
+    source=feature_view_1,
+    features=["name"],
+    keep_source_metrics=True, # enable keep_source_metrics here.
+)
+
+# This will report all metrics defined in feature_view_1 and feature_view_2. But
+# as feature_view_1 does not set keep_source_metrics to True, metrics defined in
+# feature_view_0 will not be reported.
+self.client.materialize_features(features, sink=...).wait()
+```
+
 ## Metric reporting format
 
 Metrics reported by metric stores will have the following format by default.
--- a/python/feathub/feature_tables/sinks/sink.py
+++ b/python/feathub/feature_tables/sinks/sink.py
@@ -24,6 +24,7 @@ class Sink(FeatureTable, ABC):
     Base class for all Sink Feature Table.
     """
 
+    # TODO: make sure sink subclasses expose the ability to configure table name.
     def __init__(
         self,
         name: str,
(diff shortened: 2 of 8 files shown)


python/feathub/metric_stores/metric_store.py
```diff
@@ -111,6 +116,13 @@ def create_metric_materialization_descriptors(
                     allow_overwrite=True,
                 )
             )
+        if feature_descriptor.keep_source_metrics:
+            descriptors.extend(
+                self.create_metric_materialization_descriptors(
+                    feature_descriptor=feature_descriptor.get_resolved_source(),
+                    data_sink=BlackHoleSink(),
```
Sxnan, line 123: It seems that only the sink name is used to set the "table_name" label of the Prometheus metric. Why can't we use the feature view name instead?

Sxnan (commented):
Thanks for the PR! It looks good overall. Just left one comment below.

python/feathub/feature_tables/sinks/sink.py
```diff
@@ -24,6 +24,7 @@ class Sink(FeatureTable, ABC):
     Base class for all Sink Feature Table.
     """
 
+    # TODO: make sure sink subclasses pass in meaningful table name.
```
lindong28, line 27: Can you explain the motivation of this TODO and how we can possibly achieve this?

The `name` field is given by user, similar to how user can define features' names. It is not clear how we can possibly enforce this TODO.

docs/content/metric-stores/overview.md
```diff
@@ -44,6 +44,31 @@ f_total_cost = Feature(
 )
 ```
 
+Then when the FeatureView that hosts the above feature is being materialized,
+the related metrics will also be reported during the materialization process.
+
+By default, Feathub would only report the metrics directly defined in the
+FeatureView to be materialized. If metrics defined in the source FeatureViews
+should also be reported, please enable the `keep_source_metrics` parameter when
```
lindong28, line 52: Can you update the doc to also make it clear that this will be done recursively?

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -80,6 +81,9 @@ def __init__(
                             expression is evaluated after other transformations in the
                             feature view, and only those rows which evaluate to True
                             will be outputted by the feature view.
+        :param keep_source_fields: True iff the metrics defined in the source of this
+                                   feature view should also be reported when this
```
lindong28, line 85: Can you update the doc to also make it clear that this will be done recursively?

For example, we can say something like this:

If it is true and this feature view is materialized to a sink, FeatHub will recursively enumerate source feature view of this and every upstream feature views whose keep_source_fields == true, and report metrics defined in those feature views.

python/feathub/feature_views/derived_feature_view.py
```diff
@@ -51,6 +51,7 @@ def __init__(
         features: Sequence[Union[str, Feature]],
         keep_source_fields: bool = False,
         filter_expr: Optional[str] = None,
+        keep_source_metrics: bool = False,
```
lindong28, line 54: Can you also update the python doc of `Feature#metrics` to explicitly specify that FeatHub will report metrics defined there iff the FeatureView is materialized to a sink?

lindong28 (commented):
Thanks for the PR! Left some minor comments.

python/feathub/feature_tables/sinks/sink.py
yunfengzhou-hub: Sorry for the ambiguity. Currently Sink implementations have not allowed users to specify the sink table name and would only pass an empty string to the super constructor, for example
```python

class RedisSink(Sink):
    def __init__(
        self,
         ...
    ):
        super().__init__(
            name="",
            system_name="redis",
            table_uri={
                "host": host,
                "port": port,
                "db_num": db_num,
                "namespace": namespace,
            },
            keep_timestamp_field=keep_timestamp_field,
        )
```

The goal of this TODO is to expose the ability to configure sink table name to users. I have updated the TODO to clarify this.

yunfengzhou-hub:

lindong28:
Thanks for the PR. LGTM.

PR #185, opened 2023-09-13
Add metric CountMap and Average

yunfengzhou-hub:

This PR mainly adds metrics CountMap and Average.

## Brief change log

  - Introduce built-in function `SIZE`
  - Adds metrics `CountMap` and `Average`.
  - Fixes the bug that prometheus sink function does not remove map-typed metrics with outdated labels.

  - The public API: yes

## Documentation

  - If yes, how is the feature documented? docs & python docs

### DIFF
--- a/docs/content/metric-stores/metrics.md
+++ b/docs/content/metric-stores/metrics.md
@@ -55,3 +55,58 @@ It exposes the following metric-specific tags:
 See
 [here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_prometheus_metric_store.py#L122)
 for example usages.
+
+## CountMap
+
+CountMap is a map that maps each unique feature value to the number of
+occurrences of this value. It has the following parameters:
+
+- filter_expr: Optional with None as the default value. If it is not None, it
+  represents a partial FeatHub expression which evaluates to a boolean value.
+  The partial Feathub expression should be a binary operator whose left child is
+  absent and would be filled in with the host feature name. For example, "IS
+  NULL" will be enriched into "{feature_name} IS NULL". Only features that
+  evaluate this expression into True will be considered when computing the
+  metric.
+- window_size: Optional with 0 as the default value. The time range to compute
+  the metric. It should be zero or a positive time span. If it is zero, the
+  metric will be computed from all feature values that have been processed since
+  the Feathub job is created.
+
+It exposes the following metric-specific tags:
+
+- metric_type: "count_map"
+- value: The feature value whose occurences are counted.
+- filter_expr: The value of the filter_expr parameter.
+- window_size_sec: The value of the window_size parameter in seconds.
+
+See
+[here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_metric_store.py#L87)
+for example usages.
+
+## Average
+
+Average is a metric that that shows the average of feature values. It has the
+following parameters:
+
+- filter_expr: Optional with None as the default value. If it is not None, it
+  represents a partial FeatHub expression which evaluates to a boolean value.
+  The partial Feathub expression should be a binary operator whose left child is
+  absent and would be filled in with the host feature name. For example, "IS
+  NULL" will be enriched into "{feature_name} IS NULL". Only features that
+  evaluate this expression into True will be considered when computing the
+  metric.
+- window_size: Optional with 0 as the default value. The time range to compute
+  the metric. It should be zero or a positive time span. If it is zero, the
+  metric will be computed from all feature values that have been processed since
+  the Feathub job is created.
+
+It exposes the following metric-specific tags:
+
+- metric_type: "average"
+- filter_expr: The value of the filter_expr parameter.
+- window_size_sec: The value of the window_size parameter in seconds.
+
+See
+[here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_metric_store.py#L76)
+for example usages.
--- a/java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/prometheus/MapMetricUpdater.java
+++ b/java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/prometheus/MapMetricUpdater.java
@@ -22,6 +22,7 @@ import io.prometheus.client.SimpleCollector;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Map;
 
 /** The MapMetricUpdater updates Prometheus gauge metrics with a Map from String to Double. */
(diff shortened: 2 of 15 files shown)


yunfengzhou-hub:
Hi @Sxnan could you please take a look at this PR?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/prometheus/MapMetricUpdater.java
```diff
@@ -43,24 +45,38 @@ public MapMetricUpdater(String name, Map<String, String> labels, CollectorRegist
...
-                        }
-                    }
-                    return collector.labels(labelValues.toArray(new String[0]));
-                });
+        return children.computeIfAbsent(mapKey, label -> collector.labels(getLabelValues(mapKey)));
     }
 
     @Override
     public void update(Map<String, Double> value) {
+        // TODO: save current labels to state backend and checkpoint to make sure
```
Sxnan, line 66: I don't think we need to save the current labels. After failover, the `collector` and the `children` are empty, and we just output the next given map.

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/prometheus/MapMetricUpdater.java
```diff
@@ -43,24 +45,38 @@ public MapMetricUpdater(String name, Map<String, String> labels, CollectorRegist
...
+        //  invalid gauge children can be removed after fail-over.
+        Set<String> labelsToRemove = new HashSet<>();
```
Sxnan, line 68: It may be simpler to do the following.
```
final HashSet<String> labelsToRemove = new HashSet<>(children.keySet());
labelToRemove.removeAll(value.keySet());
```

python/feathub/metric_stores/metric.py
```diff
@@ -53,17 +57,29 @@ def get_tags(self) -> OrderedDict:
...
         - feature_name: Name of a variable containing the intermediate metric value
                         generated from previously defined transformation. For the first
                         transformation, its value is the host feature name.
+
+        The function to derive data types has the following parameter:
+        - dtype: The data type of the input feature or the intermediate metric value
+                 generated from previously defined transformation.
         """
-        pass
+        return [(x, lambda dtype: None) for x in self._get_transform_functions()]
```
Sxnan, line 79: Only the PythonUdfTransform will return an output dtype. It may be more intuitive to allow PythonUdfTransform to derive the output dtype from the input dtype in the long term. 

How about adding a TODO?

python/feathub/metric_stores/tests/test_metric_store.py
```diff
@@ -0,0 +1,241 @@
...
+
+class DummyMetricStore(MetricStore):
+    def __init__(self) -> None:
+        super(DummyMetricStore, self).__init__({})
+
+    def _get_metrics_sink(self, data_sink: Sink) -> Sink:
+        raise Exception()
+
+
+class MetricStoreTest(ABC, FeathubITTestBase):
```
Sxnan, line 40: Should we name it `MetricStoreITTest`?

Sxnan (commented):
@yunfengzhou-hub Thanks for the PR! I left some comments below.

yunfengzhou-hub:
Thanks for the comments @Sxnan. I have updated the PR according to the comments.

Sxnan:
@yunfengzhou-hub Thanks for the update! LGTM. @lindong28 Could you take a look at this PR?

docs/content/metric-stores/metrics.md
```diff
@@ -55,3 +55,60 @@ It exposes the following metric-specific tags:
 See
 [here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_prometheus_metric_store.py#L122)
 for example usages.
+
+## CountMap
+
+CountMap is a group of metrics that show the number of occurrence of each
```
lindong28, line 61: How about using the following statement:

CountMap is a map that maps each unique feature value to the number of occurrences of this value.

docs/content/metric-stores/metrics.md
```diff
@@ -55,3 +55,60 @@ It exposes the following metric-specific tags:
...
+feature value. It has the following parameters:
+
+- window_size: Optional with 0 as the default value. The time range to compute
+  the metric. It should be zero or a positive time span. If it is zero, the
```
lindong28, line 65: Do we also need to support `filter_expr` for consistency with the Count metric?

docs/content/metric-stores/metrics.md
```diff
@@ -55,3 +55,60 @@ It exposes the following metric-specific tags:
...
+It exposes the following metric-specific tags:
+
+- metric_type: "count_map"
+- window_size_sec: The value of the window_size parameter in seconds.
+
+See
+[here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_metric_store.py#L82)
+for example usages.
+
+## RatioMap
```
lindong28, line 78: Do we have a use-case that needs `RatioMap`?

It seems that this metric can be easily derived from `Count` and `CountMap` in Prometheus.

docs/content/metric-stores/metrics.md
```diff
@@ -55,3 +55,60 @@ It exposes the following metric-specific tags:
...
+It exposes the following metric-specific tags:
+
+- metric_type: "ratio_map"
+- window_size_sec: The value of the window_size parameter in seconds.
+
+See
+[here](https://github.com/alibaba/feathub/blob/master/python/feathub/metric_stores/tests/test_metric_store.py#L85)
+for example usages.
+
+## AverageSize
```
lindong28, line 97: Would it be more generic to add an `average` metric that can be applied to any numerical values, and let users derive AverageSize by using a combination of `size` and `average`?

java/feathub-connectors/flink-connectors/flink-connector-prometheus/src/main/java/com/alibaba/feathub/flink/connectors/prometheus/PrometheusSinkFunction.java
```diff
@@ -193,6 +193,9 @@ public void updateMetric(RowData row, int index) throws IOException {
                 break;
             case MAP:
                 final MapData mapData = row.getMap(index);
+                if (mapData == null) {
```
lindong28, line 196: When would this happen? I am wondering if we should throw exception in this case.

python/feathub/metric_stores/metric.py
```diff
@@ -192,3 +210,150 @@ def from_json(cls, json_dict: Dict) -> "Ratio":
...
+                lambda feature_name: SlidingWindowTransform(
+                    expr=f"CAST({feature_name} AS STRING)",
+                    agg_func="VALUE_COUNTS",
+                    window_size=self.window_size,
+                    step_size=self.window_size,
+                ),
+                lambda dtype: None,
+            ),
+            (
+                lambda feature_name: PythonUdfTransform(
```
lindong28, line 305: It might be more intuitive to let `Transformation` specify its output data type, instead of making back-and-forth changes to the metric API.

And it seems simpler not to support RatioMap if it is easy for users to derive it by themselves in Prometheus.

python/feathub/processors/flink/ast_evaluator/functions.py
```diff
@@ -41,4 +42,8 @@ def evaluate_function(func_name: str, args: List[Any]) -> str:
         return f"MAP[{', '.join(args)}]"
     elif func_name == "ARRAY":
         return f"ARRAY[{', '.join(args)}]"
+    elif func_name == "SIZE":
+        if len(args) != 1:
```
lindong28, line 46: Is this already validated by the `size_type_strategy()` in `built_in_func.py`?

lindong28 (commented):

yunfengzhou-hub:

python/feathub/metric_stores/metric.py
```diff
@@ -192,3 +191,125 @@ def from_json(cls, json_dict: Dict) -> "Ratio":
...
+                            a positive time span. If it is zero, the metric will be
+                            computed from all feature values that have been processed
+                            since the Feathub job is created.
+        """
+        super().__init__("count_map", window_size)
+        self.filter_expr = filter_expr
+
+    def get_tags(self) -> OrderedDict:
+        tags = super(CountMap, self).get_tags()
+        tags["count_map_key"] = "null"
```
lindong28, line 226: Do we need to have `count_map_key` as part of this metric's tags?

Note that this tag is not documented for the CountMetric in metrics.md. 

It is mentioned in the Java doc of `PrometheusSinkFunction` that `any label whose value is "null" in the column's comment should be updated to use the given key as the label value`. Thus, I am not sure we still need to explicitly specify `count_map_key=null` anywhere in FeatHub. It might be simpler to remove this string entirely from FeatHub.

lindong28 (commented):
Thanks for the update. Left one comment below.

yunfengzhou-hub:
Thanks for the comment @lindong28. I have updated PR according to the comment.

lindong28:
Thanks for the update! LGTM.

